PR #4, opened 2022-08-17
Data preparation

JLrumberger:

It adds a skeleton to the `prepare_example` function that fleshes out how a single training example will be generated. This is done to check if anything is missing in the design concept.

The necessary functions are chained and their output is used to create a single training example. Inputs and outputs are still dummys at this stage

**Remaining issues**

None

### DIFF
--- a/cell_classification/segmentation_data_prep.py
+++ b/cell_classification/segmentation_data_prep.py
@@ -1,3 +1,4 @@
+from tkinter.messagebox import NO
 from tifffile import imread
 from skimage.morphology import erosion
 import tensorflow as tf
@@ -65,6 +66,10 @@ class SegmentationTFRecords:
             self.normalization_dict = self.calculate_normalization_matrix(
                 normalization_dict_path, normalization_quantile
             )
+            self.cell_mask_key = cell_mask_key
+            self.sample_key = sample_key
+            self.dataset = dataset
+            self.imaging_platform = imaging_platform
 
     def get_image(self, data_folder, marker):
         """Loads the images from a single data_folder
@@ -79,7 +84,7 @@ class SegmentationTFRecords:
             np.array:
                 The multiplexed image
         """
-        return None
+        return np.zeros([500, 500])
 
     def get_instance_masks(self, data_folder, cell_mask_key):
         """Makes a binary mask from an instance mask by eroding it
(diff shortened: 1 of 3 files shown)


cell_classification/segmentation_data_prep.py
```diff
@@ -62,33 +77,43 @@ def get_image(self, data_folder, marker):
...
             np.array:
                 The binary mask
+            np.array:
+                The instance mask
         """
-        pass
+        return np.zeros([500, 500]), np.zeros([500, 500])
 
-    def get_cell_types(self, labels):
+    def get_cell_types(self, labels, sample_name):
```
ngreenwald: The arguments here don't match up with the arguments down below in the example. 

I don't think we need the labeled image though, can't we just subset the `cell_table` to the name of image we're looking at to get the list of cells present in the image? Is there a reason to use the mask?

cell_classification/segmentation_data_prep.py
```diff
@@ -119,17 +143,65 @@ def marker_activity_mask(self, instance_mask, cell_types, marker_activity):
...
+                Example dict
+        """
+        # load and normalize the multiplexed image and masks
+        mplex_img = self.get_image(data_folder, marker)
+        mplex_img /= self.normalization_dict[marker]
+        binary_mask, instance_mask = self.get_instance_masks(
+            data_folder, self.cell_mask_key
+        )
+        # get the cell types and marker activity mask
+        cell_types = self.get_cell_types(instance_mask, self.sample_key)
```
ngreenwald: I think the `get_cell_types` function will need to return a df, not just a list. 

If it just returns a list, let's say `[CD8T, CD4T, CD4T, Tumor]`, and then we get the marker activity for CD8 `[1, 0, 0, -1]`, in order to generate the `marker_activity_mask`, we need to know the integer label of each cell, not just their cell type. That's how we can map back to the instance mask. 

Therefore, something like pandas df with both the cell_type and cell_ID column should probably be returned. Thsi would involve just subsetting the `cell_table` to those columns.

ngreenwald (changes requested):
Looks great, just two small changes. Let me know if I misunderstood something

cell_classification/segmentation_data_prep.py
JLrumberger: Yes, it should  be enough to subset the `cell_table` with the name of the image. I included the labels, because I thought that the images will be tiled before the next steps. But now I am thinking it'll be best if tiling is done at last in function `tile_example`.

cell_classification/segmentation_data_prep.py
JLrumberger: I agree.

ngreenwald (approved):

PR #5, opened 2022-08-18
Calculate normalization matrix

JLrumberger:

Added `calculate_normalization_matrix` and `get_instance_mask` functions and tests.

Replaced dummy code in `segmentation_data_prep.py`.

**Remaining issues**

None

### DIFF
--- a/cell_classification/segmentation_data_prep.py
+++ b/cell_classification/segmentation_data_prep.py
@@ -2,6 +2,7 @@ from tifffile import imread
 from skimage.morphology import erosion
 import tensorflow as tf
 import numpy as np
+import pandas as pd
 import os
 import xarray
 from tqdm import tqdm
@@ -58,17 +59,30 @@ class SegmentationTFRecords:
             cell_mask_key (str):
                 The key in the data_folder that contains the cell mask labels
         """
-        pass
-        if normalization_dict_path is not None:
-            self.normalization_dict = json.load(open(normalization_dict_path, "r"))
+        os.makedirs(tf_record_path, exist_ok=True)
+        # assign selected markers
+        self.conversion_matrix = pd.read_csv(conversion_matrix_path)
+        if selected_markers is None:
+            self.selected_markers = list(self.conversion_matrix.columns)
+            if cell_type_key in self.selected_markers:
+                self.selected_markers.remove(cell_type_key)
         else:
+            self.selected_markers = selected_markers
+        # load or construct normalization dict
+        if normalization_dict_path is None:
             self.normalization_dict = self.calculate_normalization_matrix(
-                normalization_dict_path, normalization_quantile
+                data_folders,
+                os.path.join(tf_record_path, "normalization_dict.json"),
+                normalization_quantile,
+                selected_markers=self.selected_markers,
             )
-            self.cell_mask_key = cell_mask_key
-            self.sample_key = sample_key
-            self.dataset = dataset
-            self.imaging_platform = imaging_platform
+        else:
+            self.normalization_dict = json.load(open(normalization_dict_path, "r"))
+
+        self.cell_mask_key = cell_mask_key
+        self.sample_key = sample_key
+        self.dataset = dataset
+        self.imaging_platform = imaging_platform
 
     def get_image(self, data_folder, marker):
         """Loads the images from a single data_folder
@@ -203,7 +217,11 @@ class SegmentationTFRecords:
         return None
 
     def calculate_normalization_matrix(
-        self, normalization_dict_path, normalization_quantile
+        self,
+        data_folders,
+        normalization_dict_path,
+        normalization_quantile,
+        selected_markers,
     ):
         """Calculates the normalization matrix for the given data if it does not exist
         Args:
@@ -215,4 +233,19 @@ class SegmentationTFRecords:
             dict:
                 The normalization dict
         """
-        return {"CD8": 0.0}
+        # iterate through the data_folders and calculate the quantiles
+        quantiles = {}
+        for data_folder in data_folders:
+            for marker in selected_markers:
+                img = self.get_image(data_folder, marker)
+                if marker not in quantiles:
+                    quantiles[marker] = []
+                quantiles[marker].append(np.quantile(img, normalization_quantile))
+        # calculate the normalization matrix
+        normalization_matrix = {}
+        for marker in selected_markers:
+            normalization_matrix[marker] = 1.0 / np.mean(quantiles[marker])
+        # save the normalization matrix
+        with open(normalization_dict_path, "w") as f:
+            json.dump(normalization_matrix, f)
+        return normalization_matrix
--- a/cell_classification/segmentation_data_prep.py
+++ b/cell_classification/segmentation_data_prep.py
@@ -100,7 +100,7 @@ class SegmentationTFRecords:
         img = imread(os.path.join(data_folder, marker + ".tiff"))
         return img
 
-    def get_instance_masks(self, data_folder, cell_mask_key):
+    def get_instance_mask(self, data_folder, cell_mask_key):
         """Makes a binary mask from an instance mask by eroding it
 
         Args:
@@ -114,7 +114,10 @@ class SegmentationTFRecords:
             np.array:
                 The instance mask
         """
-        return np.zeros([500, 500]), np.zeros([500, 500])
+        instance_mask = imread(os.path.join(data_folder, cell_mask_key + ".tiff"))
+        binary_instance_mask = erosion(instance_mask, np.ones([3, 3])) > 0
+        binary_instance_mask = binary_instance_mask.astype(instance_mask.dtype)
+        return binary_instance_mask, instance_mask
 
     def get_cell_types(self, sample_name):
         """Gets the cell types from the cell table for the given labels
@@ -175,7 +178,7 @@ class SegmentationTFRecords:
         # load and normalize the multiplexed image and masks
         mplex_img = self.get_image(data_folder, marker)
         mplex_img /= self.normalization_dict[marker]
-        binary_mask, instance_mask = self.get_instance_masks(
+        binary_mask, instance_mask = self.get_instance_mask(
             data_folder, self.cell_mask_key
         )
         # get the cell types and marker activity mask
--- a/cell_classification/segmentation_data_prep.py
+++ b/cell_classification/segmentation_data_prep.py
@@ -1,5 +1,6 @@
+from tkinter import E
 from tifffile import imread
-from skimage.morphology import erosion
+from skimage.segmentation import find_boundaries
 import tensorflow as tf
 import numpy as np
 import pandas as pd
@@ -26,7 +27,8 @@ class SegmentationTFRecords:
         normalization_quantile=0.99,
         cell_type_key="cluster_labels",
         sample_key="SampleID",
-        cell_mask_key="cell_segmentation",
+        segmentation_fname="cell_segmentation",
+        segment_label_key="labels",
     ):
         """Initializes SegmentationTFRecords and loads everything except the images
 
@@ -56,33 +58,23 @@ class SegmentationTFRecords:
                 The key in the cell table that contains the cell type labels
             sample_key (str):
                 The key in the cell table that contains the sample name
-            cell_mask_key (str):
-                The key in the data_folder that contains the cell mask labels
+            segmentation_fname (str):
+                The filename in the data_folder that contains the cell instance segmentation
+            segment_label_key (str):
+                The key in the cell_table.csv that contains the cell segment labels
         """
-        os.makedirs(tf_record_path, exist_ok=True)
-        # assign selected markers
-        self.conversion_matrix = pd.read_csv(conversion_matrix_path)
-        if selected_markers is None:
-            self.selected_markers = list(self.conversion_matrix.columns)
-            if cell_type_key in self.selected_markers:
-                self.selected_markers.remove(cell_type_key)
-        else:
-            self.selected_markers = selected_markers
-        # load or construct normalization dict
-        if normalization_dict_path is None:
-            self.normalization_dict = self.calculate_normalization_matrix(
-                data_folders,
-                os.path.join(tf_record_path, "normalization_dict.json"),
-                normalization_quantile,
-                selected_markers=self.selected_markers,
-            )
-        else:
-            self.normalization_dict = json.load(open(normalization_dict_path, "r"))
-
-        self.cell_mask_key = cell_mask_key
+        self.selected_markers = selected_markers
+        self.data_folders = data_folders
+        self.normalization_dict_path = normalization_dict_path
+        self.conversion_matrix_path = conversion_matrix_path
+        self.normalization_quantile = normalization_quantile
+        self.segmentation_fname = segmentation_fname
+        self.segment_label_key = segment_label_key
         self.sample_key = sample_key
         self.dataset = dataset
         self.imaging_platform = imaging_platform
+        self.tf_record_path = tf_record_path
+        self.cell_type_key = cell_type_key
 
     def get_image(self, data_folder, marker):
         """Loads the images from a single data_folder
@@ -100,7 +92,7 @@ class SegmentationTFRecords:
         img = imread(os.path.join(data_folder, marker + ".tiff"))
         return img
 
-    def get_instance_mask(self, data_folder, cell_mask_key):
+    def get_instance_mask(self, data_folder, segmentation_fname):
         """Makes a binary mask from an instance mask by eroding it
 
         Args:
@@ -114,10 +106,10 @@ class SegmentationTFRecords:
             np.array:
                 The instance mask
         """
-        instance_mask = imread(os.path.join(data_folder, cell_mask_key + ".tiff"))
-        binary_instance_mask = erosion(instance_mask, np.ones([3, 3])) > 0
-        binary_instance_mask = binary_instance_mask.astype(instance_mask.dtype)
-        return binary_instance_mask, instance_mask
+        instance_mask = imread(os.path.join(data_folder, segmentation_fname + ".tiff"))
+        edge = find_boundaries(instance_mask, mode="inner").astype(np.uint8)
+        interior = np.logical_and(edge == 0, instance_mask > 0).astype(np.uint8)
+        return interior, instance_mask
 
     def get_cell_types(self, sample_name):
         """Gets the cell types from the cell table for the given labels
@@ -179,7 +171,7 @@ class SegmentationTFRecords:
         mplex_img = self.get_image(data_folder, marker)
         mplex_img /= self.normalization_dict[marker]
         binary_mask, instance_mask = self.get_instance_mask(
-            data_folder, self.cell_mask_key
+            data_folder, self.segmentation_fname
         )
         # get the cell types and marker activity mask
         cell_types = self.get_cell_types(data_folder)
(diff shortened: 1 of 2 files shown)


cell_classification/segmentation_data_prep.py
```diff
@@ -58,17 +59,30 @@ def __init__(
...
         """
-        pass
-        if normalization_dict_path is not None:
-            self.normalization_dict = json.load(open(normalization_dict_path, "r"))
+        os.makedirs(tf_record_path, exist_ok=True)
+        # assign selected markers
+        self.conversion_matrix = pd.read_csv(conversion_matrix_path)
+        if selected_markers is None:
+            self.selected_markers = list(self.conversion_matrix.columns)
+            if cell_type_key in self.selected_markers:
```
ngreenwald: This is fine for now, once we have a dedicated input validation function we can move that logic there instead

cell_classification/segmentation_data_prep.py
```diff
@@ -100,7 +114,10 @@ def get_instance_masks(self, data_folder, cell_mask_key):
             np.array:
                 The instance mask
         """
-        return np.zeros([500, 500]), np.zeros([500, 500])
+        instance_mask = imread(os.path.join(data_folder, cell_mask_key + ".tiff"))
```
ngreenwald: It's generally the case with our data that the cell_mask_key is not the same as the image name for the segmentation mask. I would add another argument that's specific for the segmentation_mask_name

cell_classification/segmentation_data_prep.py
```diff
@@ -100,7 +114,10 @@ def get_instance_masks(self, data_folder, cell_mask_key):
...
+        binary_instance_mask = erosion(instance_mask, np.ones([3, 3])) > 0
```
ngreenwald: This erosion doesn't work for objects that are flush with one another

```
xx = np.zeros((20, 20))
xx[1:5, 1:5] = 1
xx[5:10, 1:5] = 2
xx[1:5, 5:10] = 3
xx[5:10, 5:10] = 4
xx[10:15, 5:10] = 5
yy = erosion(xx, np.ones([3, 3]))
```

This is what we wrote for Mesmer:  https://github.com/vanvalenlab/deepcell-tf/blob/36b68e4f0a44b37067ced043d1f943d9fda5ef8f/deepcell/utils/transform_utils.py#L80

cell_classification/segmentation_data_prep.py
```diff
@@ -203,7 +220,11 @@ def make_tf_record(self, data_folders, tf_record_path):
         return None
 
     def calculate_normalization_matrix(
-        self, normalization_dict_path, normalization_quantile
+        self,
+        data_folders,
+        normalization_dict_path,
```
ngreenwald: I think it looks nicer to have the args to the function on a smaller number of lines


```
def calculate_normalization_matrix(self, data_folder,s normalization_dict_path, ... etc)
```

cell_classification/segmentation_data_prep_test.py
```diff
@@ -1,24 +1,77 @@
...
+            normalization_dict_path=norm_dict_path,
+        )
     return data_prep
 
 
+def prepare_conversion_matrix():
+    col_names = [
+        "cluster_labels",
+        "CD11c",
+        "CD14",
```
ngreenwald: For testing, it'll make your life much easier if you go with the minimal example. Probably 3-4 markers is enough

cell_classification/segmentation_data_prep_test.py
```diff
@@ -1,24 +1,77 @@
...
+        "CD45RO",
+        "CD86",
+    ]
+    row_names = [
+        "stromal",
+        "FAP",
+        "NK",
+        "CD4T",
+        "CD14",
+        "CD163",
```
ngreenwald: Same with cell types

cell_classification/segmentation_data_prep_test.py
```diff
@@ -33,9 +86,143 @@ def test_get_image():
...
+        # check if the normalization_dict has the correct values
+        for marker in norm_dict.keys():
+            assert norm_dict[marker] == 1.0
+        # check if the normalization_dict has the correct keys
+        for marker in selected_markers:
+            assert marker in norm_dict.keys()
+        # check if the normalization_dict works correctly when
+        # normalization_dict_path is given to init
+        conversion_matrix = prepare_conversion_matrix()
+        conversion_matrix_path = os.path.join(temp_dir, "conversion_matrix.csv")
```
ngreenwald, line 114: There's a lot of code here, and above, for making the `conversion_matrix`. However, this conversion matrix isn't needed for the function being tested here, only to initialize the class. This is a hint that the logic of things are being set up could be improved. 

Either 1) Create a dummy version of the SegmentationTFRecords object that can be initialized without requiring all of the arguments to be specified, to use just for testing, or 
2) Move the logic that is currently being executed in the `init` portion of the class into a dedicated helper function that is called later, so it doesn't need to be run every time

cell_classification/segmentation_data_prep_test.py
```diff
@@ -33,9 +86,143 @@ def test_get_image():
...
+        )
+        assert np.array_equal(loaded_img, instance_mask)
+        # check if binary mask is binarized correctly
+        assert np.min(loaded_binary_img) == 0
+        assert np.max(loaded_binary_img) == 1
+        # check if binary mask is eroded correctly
+        loaded_eroded_img, loaded_img = data_prep.get_instance_mask(
+            data_folder=temp_dir, cell_mask_key="cell_segmentation_full"
+        )
+        assert np.array_equal(loaded_eroded_img, instance_mask_eroded)
```
ngreenwald: Should update these tests to make the same assertion for when cells are touching

ngreenwald (changes requested):
Awesome start! For tests, it makes them easier to read if you visually separate them when you re-call the function under test with different settings.

cell_classification/segmentation_data_prep.py
JLrumberger: I moved loading and calculating the `normalization_matrix` from `__init__` to `make_tf_record` function and added `test_make_tf_record` accordingly.

cell_classification/segmentation_data_prep_test.py
JLrumberger: Put stuff form `init` into `make_tf_record` because it's needed there but not in `init`

cell_classification/segmentation_data_prep.py
```diff
@@ -161,8 +170,8 @@ def prepare_example(self, data_folder, marker):
         # load and normalize the multiplexed image and masks
         mplex_img = self.get_image(data_folder, marker)
         mplex_img /= self.normalization_dict[marker]
-        binary_mask, instance_mask = self.get_instance_masks(
-            data_folder, self.cell_mask_key
+        binary_mask, instance_mask = self.get_instance_mask(
+            data_folder, self.segmentation_fname
```
ngreenwald: No need to pass the `self.segmentation_fname` to the function, can just access it within the function right?

cell_classification/segmentation_data_prep.py
ngreenwald: These still need to be combined together into smaller number of lines

cell_classification/segmentation_data_prep.py
```diff
@@ -200,10 +209,38 @@ def make_tf_record(self, data_folders, tf_record_path):
             tf_record_path (str):
                 The path to the tf record to make
         """
+        # make tfrecord path
+        os.makedirs(tf_record_path, exist_ok=True)
+        # read conversion matrix
+        self.conversion_matrix = pd.read_csv(self.conversion_matrix_path)
+
+        # check if markers were selected or take all markers from conversion matrix
```
ngreenwald, line 200: This is great, much easier to read with some additional spacing

cell_classification/segmentation_data_prep_test.py
```diff
@@ -1,24 +1,40 @@
...
         tile_size=[256, 256],
-        tf_record_path="path_to_tf_record",
+        tf_record_path="tf_record_path",
+        normalization_dict_path="path_to_normalization_dict",
     )
     return data_prep
 
 
+def prepare_conversion_matrix():
+    col_names = ["cluster_labels", "CD11c", "CD14", "CD56", "CD57"]
```
ngreenwald: There's a couple places in the code where you have this logic: checking the column names to remove one that is really a label for the rows. Is there a reason to have this extra column name here?

```
cols = ['CD11c', 'CD4', 'CD8']
rows = ['Tcell', 'Bcell', 'tumor']
data = np.random.randint(0, 2, size=(len(rows), len(cols)))
df = pd.DataFrame(data, columns=cols, index=rows)
```

This will mean we can always assume all column names are valid

cell_classification/segmentation_data_prep_test.py
```diff
@@ -33,9 +49,165 @@ def test_get_image():
...
+                img,
+            )
+    return data_folders
+
+
+def test_calculate_normalization_matrix():
+
+    # instantiate data_prep, conversion_matrix and markers
+    data_prep = prep_object()
+    conversion_matrix = prepare_conversion_matrix()
```
ngreenwald: Now that we don't need the conversion matrix, the code will be easier to understand if you just create the list of markers directly, rather than taking them from a function

cell_classification/segmentation_data_prep_test.py
```diff
@@ -33,9 +49,165 @@ def test_get_image():
...
+            norm_dict = data_prep.calculate_normalization_matrix(
+                data_folders=data_folders,
+                normalization_dict_path=os.path.join(temp_dir, "norm_dict_test.json"),
+                normalization_quantile=0.99,
+                selected_markers=selected_markers,
+            )
+
+            # check if the normalization_dict has the correct values for stochastic images
+            for marker in norm_dict.keys():
+                assert norm_dict[marker] - 1 / (0.5 * scale) < 0.001
```
ngreenwald: If the normalization_dict stores the value of the 99.9th percentile, rather than 1/value, we can directly check that it's correct

cell_classification/segmentation_data_prep_test.py
```diff
@@ -33,9 +49,165 @@ def test_get_image():
...
+                5, temp_dir, selected_markers, random=True, scale=scale
+            )
+            norm_dict = data_prep.calculate_normalization_matrix(
+                data_folders=data_folders,
+                normalization_dict_path=os.path.join(temp_dir, "norm_dict_test.json"),
+                normalization_quantile=0.99,
+                selected_markers=selected_markers,
+            )
+
+            # check if the normalization_dict has the correct values for stochastic images
```
ngreenwald: Because the normalization value for each marker here is the same, this test would pass if the function had been written to take whatever normalization quantile was found for the first marker, and apply it to all of the other markers. 

We should check that when different markers have different normalization values, the function accurately returns them

cell_classification/segmentation_data_prep_test.py
```diff
@@ -33,9 +49,165 @@ def test_get_image():
...
+def test_make_tf_record():
+    data_prep = prep_object()
+    conversion_matrix = prepare_conversion_matrix()
+    selected_markers = list(conversion_matrix.columns)
+    selected_markers.remove("cluster_labels")
+    with tempfile.TemporaryDirectory() as temp_dir:
+
+        # create temporary folders with data and prepare normalization_dict
+        data_folders = prepare_test_data_folders(5, temp_dir, selected_markers)
+        norm_dict = data_prep.calculate_normalization_matrix(
```
ngreenwald: This is what's called an integration test. We already know that `calculate_normalization_matrix` works, because it has its own test function . So we don't need to recheck that the output is correct for every function that calls `calculate_normalization_matrix`. Instead, what we need to do is make sure the functions work together as expected. 

Therefore, calling `calculate_normalization_matrix` and checking the output against `data_prep.normalization_dict` is overkill. Instead, all we need to do is make sure that *whatever* normalization dict is loaded, that *same* normalization dict can be found added as an attribute. This allow us to further separate the tests from one another, and makes it shorter. 

I would make a simple normalization dict with two entries, save it, and make sure that same one is loaded.

cell_classification/segmentation_data_prep_test.py
```diff
@@ -33,9 +49,165 @@ def test_get_image():
...
+            assert norm_dict_loaded == norm_dict
+
+            # check if the normalization_dict has the correct keys
+            for marker in selected_markers:
+                assert marker in norm_dict.keys()
+
+
+def test_make_tf_record():
+    data_prep = prep_object()
+    conversion_matrix = prepare_conversion_matrix()
```
ngreenwald: This code is found here, and also 15 lines down from here on line 129. We don't need both of them, as its redundant.

cell_classification/segmentation_data_prep_test.py
```diff
@@ -33,9 +49,165 @@ def test_get_image():
...
+
+        # check if binary mask is eroded correctly
+        assert np.array_equal(loaded_binary_img, instance_mask_eroded)
+
+
 def test_prepare_example():
     data_prep = prep_object()
     with tempfile.TemporaryDirectory() as temp_dir:
+        data_folders = prepare_test_data_folders(5, temp_dir, selected_markers=["CD8"])
+        norm_dict = data_prep.calculate_normalization_matrix(
```
ngreenwald: Same thing here. We don't need to run the actual `calculate_normalization_matrix` function for this integration test. We just want any placeholder normalization dict to be attached to `data_prep`. In fact, it's often advantageous to make a placeholder normalization_dict ourselves in the test function, because then it's very obvious what the values are.

cell_classification/segmentation_data_prep.py
```diff
@@ -215,4 +252,22 @@ def calculate_normalization_matrix(
...
+        # iterate through the data_folders and calculate the quantiles
+        quantiles = {}
+        for data_folder in data_folders:
+            for marker in selected_markers:
+                img = self.get_image(data_folder, marker)
+                if marker not in quantiles:
+                    quantiles[marker] = []
+                quantiles[marker].append(np.quantile(img, normalization_quantile))
+
+        # calculate the normalization matrix
```
ngreenwald, line 246: Same here, much easier to read

cell_classification/segmentation_data_prep_test.py
```diff
@@ -33,9 +49,165 @@ def test_get_image():
         assert not np.array_equal(CD8_img, CD4_img)
 
 
+def prepare_test_data_folders(
+    num_folders, temp_dir, selected_markers, random=False, scale=1.0
```
ngreenwald: These can be moved up a line

ngreenwald (changes requested):
Looks great! All of the tests do their job. Now the goal is to have them do their job, but in as simple and easy to understand a way as possible.

The main idea behind integration testing is that when you're testing `function_x`, you make sure the output of `function_x` is correct. However, when `function_y` calls `function_x` as part of its own internal logic, you don't need to check, again, that the output of `function_x` is correct, since you already checked it. Instead, you just need to make sure that `function_x`  and `function_y` integrate together successfully.

cell_classification/segmentation_data_prep_test.py
JLrumberger: It's just how I imagined the `conversion_matrix` might look like when set up in excel. I can get rid of the extra column name and store the labels in the row names if this is desired.

JLrumberger:
I think I implemented all requested changes. My code formatter (black) seems to insert line breaks for init arguments.. so this might happen again in the future when I forget to fix it after formatting.

cell_classification/segmentation_data_prep.py
```diff
@@ -100,7 +97,12 @@ def get_instance_masks(self, data_folder, cell_mask_key):
             np.array:
                 The instance mask
         """
-        return np.zeros([500, 500]), np.zeros([500, 500])
+        instance_mask = imread(
+            os.path.join(data_folder, self.segmentation_fname + ".tiff")
```
ngreenwald: Your default formatter probably has 60 characters as the limit, but we use 99 characters in the repo as our limit. See if you can adjust it so that when it autoformats it doesn't try and fit everything in only 60

cell_classification/segmentation_data_prep_test.py
```diff
@@ -1,24 +1,44 @@
...
         tile_size=[256, 256],
-        tf_record_path="path_to_tf_record",
+        tf_record_path=tf_record_path,
+        normalization_dict_path=normalization_dict_path,
     )
     return data_prep
 
 
+def prepare_conversion_matrix():
+    col_names = ["CD11c", "CD14", "CD56", "CD57"]
```
ngreenwald, line 32: Much better!

cell_classification/segmentation_data_prep.py
```diff
@@ -86,7 +83,7 @@ def get_image(self, data_folder, marker):
         img = imread(os.path.join(data_folder, marker + ".tiff"))
         return img
 
-    def get_instance_masks(self, data_folder, cell_mask_key):
+    def get_instance_mask(self, data_folder):
```
ngreenwald: I would give this function a different name, since it returns both the instance and binary mask

cell_classification/segmentation_data_prep.py
```diff
@@ -192,19 +192,42 @@ def tile_example(example, tile_size):
...
+        """Checks the input for correctness"""
+        # make tfrecord path
+        os.makedirs(self.tf_record_path, exist_ok=True)
+        # read conversion matrix
+        self.conversion_matrix = pd.read_csv(self.conversion_matrix_path)
+
+        # check if markers were selected or take all markers from conversion matrix
+        if self.selected_markers is None:
+            self.selected_markers = list(self.conversion_matrix.columns)
+            if self.cell_type_key in self.selected_markers:
```
ngreenwald: Don't need this line anymore

cell_classification/segmentation_data_prep_test.py
```diff
@@ -33,9 +53,122 @@ def test_get_image():
         assert not np.array_equal(CD8_img, CD4_img)
 
 
-def test_prepare_example():
+def prepare_test_data_folders(num_folders, temp_dir, selected_markers, random=False, scale=1.0):
```
ngreenwald: If you make `scale` a list, or have `scale` linearly increase for each `marker` in `selected_markers`, you'll be able to test different normalization coefficients for different markers

cell_classification/segmentation_data_prep_test.py
```diff
@@ -33,9 +53,122 @@ def test_get_image():
...
+
+    # check normalization_dict for different stochastic images
+    for scale in [0.5, 9.132]:
+
+        # create temporary folders with data and do tests
+        with tempfile.TemporaryDirectory() as temp_dir:
+            data_folders = prepare_test_data_folders(
+                5, temp_dir, selected_markers, random=True, scale=scale
+            )
+            data_prep = prep_object(
```
ngreenwald: This is repeated from line 77. I think you only need this one

cell_classification/segmentation_data_prep_test.py
ngreenwald: This one still needs to be addressed

cell_classification/segmentation_data_prep_test.py
ngreenwald: You can decide if you want this one or not

ngreenwald (changes requested):
Looks great, just a couple tweaks. 

For PRs that have a lot of comments, I recommend clicking the "Resolve conversation" button after you've addressed an individual comment, that way it's easy to see what's left. 
image

cell_classification/segmentation_data_prep.py
JLrumberger: Renamed to `get_inst_binary_masks`

cell_classification/segmentation_data_prep.py
JLrumberger: It had 88 and I changed it to 99. Thanks for the suggestion. It now only adds excessive line breaks for arguments in function definitions.

ngreenwald (approved):
Awesome job!

PR #7, opened 2022-08-22
Make marker activity GT masks

JLrumberger:

Constructing the gt marker activity masks given the cell segmentations, the conversion table and the cell table.

I implemented the functions 
- `get_cell_types(sample_name)`: subsets `cell_table` and returns only the rows that come from the same FOV (i.e. same sample_name)
- `get_marker_activity(cell_types, conversion_matrix, markers)`: returns a dictionary with marker names as keys and lists of cells as values. The lists are binary and indicate if a given cell is marker positive or not
- `get_marker_activity_mask(instance_mask, binary_mask, marker_activity, markers)`: maps the marker activity onto the instance segmentation maps to return a spatial representation of positive/negative marker activity for a list of given markers.

**Remaining issues**

Some small things here and there could be done better. I am not happy with the marker activity dict. It looks like this
`{"marker":[1,0,0,0,1]}` where the first element of the list corresponds to marker activity in the cell segment with label 1. I think this list should be pre-pended with a zero so that the indexes of the list correspond to the cell labels. Anyways, I am interested in your comments and I guess you'll find more weaknesses.

### DIFF
--- a/cell_classification/segmentation_data_prep.py
+++ b/cell_classification/segmentation_data_prep.py
@@ -66,6 +66,7 @@ class SegmentationTFRecords:
         self.imaging_platform = imaging_platform
         self.tf_record_path = tf_record_path
         self.cell_type_key = cell_type_key
+        self.cell_table_path = cell_table_path
 
     def get_image(self, data_folder, marker):
         """Loads the images from a single data_folder
@@ -112,7 +113,8 @@ class SegmentationTFRecords:
             df:
                 The labels and corresponding cell types for the given sample
         """
-        return None
+
+        return self.cell_type_table[self.cell_type_table.SampleID == sample_name]
 
     def get_marker_activity(self, cell_types, marker):
         """Gets the marker activity for the given labels
@@ -211,6 +213,9 @@ class SegmentationTFRecords:
                 self.data_folders,
                 self.selected_markers,
             )
+
+        # load cell_types.csv
+        self.cell_type_table = pd.read_csv(self.cell_table_path)
 
     def make_tf_record(self, data_folders):
         """Iterates through the data_folders and loads, transforms and
--- a/cell_classification/segmentation_data_prep.py
+++ b/cell_classification/segmentation_data_prep.py
@@ -103,66 +103,43 @@ class SegmentationTFRecords:
         interior = np.logical_and(edge == 0, instance_mask > 0).astype(np.uint8)
         return interior, instance_mask
 
-    def get_cell_types(self, sample_name):
-        """Gets the cell types from the cell table for the given labels
-        Args:
-            sample_name (str):
-                The name of the sample we use for look up in column sample_key
-                in cell_type.csv
-        Returns:
-            df:
-                The labels and corresponding cell types for the given sample
-        """
-
-        return self.cell_type_table[self.cell_type_table.SampleID == sample_name]
-
-    def get_marker_activity(self, cell_types, conversion_matrix, markers):
+    def get_marker_activity(self, sample_name, conversion_matrix, marker):
         """Gets the marker activity for the given labels
         Args:
-            cell_types array:
-                The cell types to get the marker activity for
-            marker (str, list):
-                The markers to get the activity for
+            sample_name (str):
+                The name of the sample
             conversion_matrix (pd.DataFrame):
                 The conversion matrix to use for the lookup
+            marker (str, list):
+                The markers to get the activity for
         Returns:
             np.array:
                 The marker activity for the given labels, 1 if the marker is active, 0
                 otherwise and -1 if the marker is not specific enough to be considered active
         """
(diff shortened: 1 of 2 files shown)


cell_classification/segmentation_data_prep.py
```diff
@@ -112,23 +113,33 @@ def get_cell_types(self, sample_name):
             df:
                 The labels and corresponding cell types for the given sample
         """
-        return None
 
-    def get_marker_activity(self, cell_types, marker):
+        return self.cell_type_table[self.cell_type_table.SampleID == sample_name]
```
ngreenwald: This probably doesn't need to be a whole function. We can decide later if we want to remove it, fine for now

cell_classification/segmentation_data_prep.py
```diff
@@ -142,7 +153,16 @@ def get_marker_activity_mask(self, instance_mask, cell_types, marker_activity):
             np.array:
                 The marker activity mask
         """
-        return np.zeros([500, 500])
+        out_list = []
+        if isinstance(markers, str):
+            markers = [markers]
+        for marker in markers:
+            out_mask = np.zeros_like(instance_mask)
+            for label, activity in enumerate(marker_activity[marker], 1):
```
ngreenwald: I think we can combine these two functions together, and have them handle only a single marker at a time. 

Then the input will be the conversion matrix, cell types, instance mask, and a single marker, and the output will be the out_mask for that single marker. 

This removes the need for a lot of the intermediate data structures. We can then either stack or concatenate the resulting the masks depending on how many channels we want the network to see at a time

ngreenwald (changes requested):
Looks great, super clean code

cell_classification/segmentation_data_prep.py
JLrumberger: I merged `get_cell_types` into `get_marker_activity` so that this function handles all the tabular data lookups and returns the marker activity for a given fov and a given marker. Thus `get_marker_activity_mask` will just take instance and binary masks and the marker activity.

cell_classification/segmentation_data_prep.py
```diff
@@ -102,47 +103,43 @@ def get_inst_binary_masks(self, data_folder):
...
+                The binary mask to make the marker activity mask for
             marker_activity list:
                 The marker activity that maps to the cell types
         Returns:
             np.array:
                 The marker activity mask
         """
-        return np.zeros([500, 500])
+        out_mask = np.zeros_like(instance_mask)
+        for label, activity in enumerate(marker_activity, 1):
```
ngreenwald: We don't know that the segmentation masks are going to be sequentially labeled without missing values. 
The post-processing we use for Mesmer, for example, will remove cells below a certain threshold, but it won't relabel the mask. This means it could go [1, 2, 4, 5, 6, 8, etc..]. We should probably directly iterate over the labels, rather than assuming that the index is equivalent to the labels.

cell_classification/segmentation_data_prep_test.py
```diff
@@ -168,5 +186,55 @@ def test_get_inst_binary_masks():
...
+
+    marker_activity_mask = data_prep.get_marker_activity_mask(
+        instance_mask, binary_mask, marker_activity
+    )
+
+    # check if the right spatial dimensions got returned
+    assert marker_activity_mask.shape == instance_mask.shape
+
+    # check if the right marker activity values are returned
+    instance_mask[binary_mask < 1] = 0
```
ngreenwald: What is the point of this line?

cell_classification/segmentation_data_prep_test.py
```diff
@@ -168,5 +186,55 @@ def test_get_inst_binary_masks():
...
+    # check if the out_dict has the right marker activity values for a given cell
+    for i in range(len(fov_1_subset.label)):
+        assert (
+            marker_activity[i] == conversion_matrix.loc[fov_1_subset.cluster_labels[i], "CD11c"]
+        )
+
+
+def test_get_marker_activity_mask():
+    data_prep = prep_object()
+    marker = "CD11c"
```
ngreenwald: This isn't used, right?

ngreenwald (changes requested):
Looks great, just a couple small comments

cell_classification/segmentation_data_prep_test.py
JLrumberger: Was there because I tested the function with the real eroded binary mask in the first place. I changed this to a simple non-eroded binary mask and thus don't need this line anymore.

cell_classification/segmentation_data_prep_test.py
JLrumberger: No, the marker isn't used anymore.

ngreenwald (approved):
Awesome stuff

PR #9, opened 2022-08-23
Image augmentation pipeline

### LINKED ISSUE #8: Image augmentation pipeline
**Section 1: Design details**

We want to have an image augmentation pipeline that randomly transforms the images. PR #9 

**Relevant background**

DL models are known to become more robust and overfit less when augmentations are used during training. Thus we want to have an augmentation pipeline that applies spatial and intensity augmentations onto images and solely spatial augmentations on the masks.

**Design overview**

- `get_augmentation_pipeline(params)` function that takes in a parameter dict and returns an object of class `imagaug.augmenters.Sequential`.
- `augment_images(images, masks, augmentation_pipeline)` function that takes in the augmentation pipeline, input images and masks and returns augmented images and masks

**Design list/flowchart**

- `get_augmentation_pipeline(params)` produces pipeline object
- `augment_images(images, masks, augmentation_pipeline)` applies augmentation pipeline onto our data. For images bilinear interpolation is used, for masks nearest-neighbor interpolation so that their unique values don't change.

**Required inputs**

- `params = {
        # flip
        "flip_prob": 1.0,
        # affine
        "affine_prob": 1.0, "scale_min": 0.5, "scale_max": 1.5, "shear_angle": 10,
        # elastic
        "elastic_prob": 1.0, "elastic_alpha": [0, 5.0], "elastic_sigma": 0.5,
        # rotate
        "rotate_count": [0, 3],
        # gaussian noise
        "gaussian_noise_prob": 1.0, "gaussian_noise_min": 0.1, "gaussian_noise_max": 0.5,
        # gaussian blur
        "gaussian_blur_prob": 0.0, "gaussian_blur_min": 0.1, "gaussian_blur_max": 0.5,
        # contrast aug
        "contrast_prob": 0.0, "contrast_min": 0.1, "contrast_max": 2.0,
    }`
- images (np.array, dtype=float32) B x H x W x C
- masks (np.array, dtype=uint8/16) B x H x W x C

**Output files**

Intermediate output: augmentation pipeline object of class `imagaug.augmenters.Sequential`
Final output: augmented images (np.array, dtype=float32) B x H x W x C and masks (np.array, dtype=uint8/16) B x H x W x C

**Control flow**

Images need to be interpolated with bilinear interpolation, masks need to be interpolated with nearest-neighbor interpolation, otherwise the they will contain label values that weren't there before.

**Milestones and timeline**

Ready for review.

JLrumberger:

Adding an image augmentation pipeline as closes #8 .

With `imgaug` library. We have one function that constructs the image augmentation pipeline and another function that takes the pipeline and the data and transforms it.

**Remaining issues**

None

### DIFF
--- a/cell_classification/augmentation_pipeline.py
+++ b/cell_classification/augmentation_pipeline.py
@@ -24,8 +24,8 @@ def augment_images(images, masks, augmentation_pipeline):
     masks = [SegmentationMapsOnImage(mask, shape=images.shape[1:-1]) for mask in masks]
     batch = Batch(images=images, segmentation_maps=masks)
     batch = augmentation_pipeline.augment_batch_(batch)
-    augmented_masks = [np.squeeze(mask.arr, -1) for mask in batch.segmentation_maps_aug]
-    return np.stack(batch.images_aug, 0), np.stack(augmented_masks, 0)
+    augmented_masks = [np.squeeze(mask.arr) for mask in batch.segmentation_maps_aug]
+    return np.stack(batch.images_aug, 0), np.squeeze(np.stack(augmented_masks, 0))
 
 
 def get_augmentation_pipeline(params):
--- a/cell_classification/augmentation_pipeline.py
+++ b/cell_classification/augmentation_pipeline.py
@@ -21,11 +21,16 @@ def augment_images(images, masks, augmentation_pipeline):
         np.array:
             The augmented masks
     """
-    masks = [SegmentationMapsOnImage(mask, shape=images.shape[1:-1]) for mask in masks]
+    masks = [SegmentationMapsOnImage(mask, shape=mask.shape) for mask in masks]
     batch = Batch(images=images, segmentation_maps=masks)
     batch = augmentation_pipeline.augment_batch_(batch)
-    augmented_masks = [np.squeeze(mask.arr) for mask in batch.segmentation_maps_aug]
-    return np.stack(batch.images_aug, 0), np.squeeze(np.stack(augmented_masks, 0))
+    augmented_masks = [mask.arr for mask in batch.segmentation_maps_aug]
+    augmented_masks = np.stack(augmented_masks, 0)
+
+    # remove additional channel from single channel masks
+    if augmented_masks.shape[-1] == 1:
+        augmented_masks = np.squeeze(augmented_masks, -1)
+    return np.stack(batch.images_aug, 0), augmented_masks
 
 
 def get_augmentation_pipeline(params):
@@ -40,8 +45,10 @@ def get_augmentation_pipeline(params):
     """
     augmentation_pipeline = iaa.Sequential(
         [
+            # random mirroring along horizontal and vertical axis
             iaa.Fliplr(params["flip_prob"]),
             iaa.Flipud(params["flip_prob"]),
+            # random zooming and shearing
             iaa.Sometimes(
                 params["affine_prob"],
                 iaa.Affine(
@@ -49,13 +56,16 @@ def get_augmentation_pipeline(params):
                     shear=(-params["shear_angle"], params["shear_angle"]),
                 ),
             ),
+            # elastic transformations that apply a water-like effect onto the image
             iaa.Sometimes(
                 params["elastic_prob"],
                 iaa.ElasticTransformation(
                     alpha=params["elastic_alpha"], sigma=params["elastic_sigma"]
                 ),
             ),
+            # 90 degree rotations
             iaa.Rot90(params["rotate_count"]),
+            # random gaussian noise added to the image
             iaa.Sometimes(
                 params["gaussian_noise_prob"],
                 iaa.AdditiveGaussianNoise(
@@ -63,12 +73,14 @@ def get_augmentation_pipeline(params):
                     scale=(params["gaussian_noise_min"], params["gaussian_noise_max"]),
                 ),
             ),
+            # random blurring with a gaussian filter
             iaa.Sometimes(
                 params["gaussian_blur_prob"],
                 iaa.GaussianBlur(
                     sigma=(params["gaussian_blur_min"], params["gaussian_blur_max"]),
                 ),
             ),
+            # random up-scaling and down-scaling of the image intensities
             iaa.Sometimes(
                 params["contrast_prob"],
                 iaa.LinearContrast(
--- a/cell_classification/augmentation_pipeline.py
+++ b/cell_classification/augmentation_pipeline.py
@@ -1,0 +1,80 @@
+import imgaug as ia
+import imgaug.augmenters as iaa
+from imgaug.augmentables.segmaps import SegmentationMapsOnImage
+from imgaug.augmentables.batches import Batch
+import numpy as np
+
+
+def augment_images(images, masks, augmentation_pipeline):
+    """
+    Augment images and masks.
+    Args:
+        images (np.array):
+            The images (b,h,w,c) to augment
+        masks (np.array):
+            The masks (b,h,w,1) to augment
+        augmentation_pipeline (imgaug.augmenters.meta.Augmenter):
+            The augmentation pipeline
+    Returns:
+        np.array:
+            The augmented images
+        np.array:
+            The augmented masks
+    """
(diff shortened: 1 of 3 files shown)


cell_classification/augmentation_pipeline.py
```diff
@@ -0,0 +1,80 @@
...
+    Returns:
+        np.array:
+            The augmented images
+        np.array:
+            The augmented masks
+    """
+    masks = [SegmentationMapsOnImage(mask, shape=images.shape[1:-1]) for mask in masks]
+    batch = Batch(images=images, segmentation_maps=masks)
+    batch = augmentation_pipeline.augment_batch_(batch)
+    augmented_masks = [np.squeeze(mask.arr) for mask in batch.segmentation_maps_aug]
```
ngreenwald: Good to add a comment about why the shape is changing

cell_classification/augmentation_pipeline.py
```diff
@@ -0,0 +1,80 @@
...
+    return np.stack(batch.images_aug, 0), np.squeeze(np.stack(augmented_masks, 0))
```
ngreenwald: How come we squeeze the `augmented_masks` twice?

cell_classification/augmentation_pipeline.py
```diff
@@ -0,0 +1,80 @@
...
+    Args:
+        params (dict):
+            The parameters for the augmentation
+    Returns:
+        imgaug.augmenters.meta.Augmenter:
+            The augmentation pipeline
+    """
+    augmentation_pipeline = iaa.Sequential(
+        [
+            iaa.Fliplr(params["flip_prob"]),
```
ngreenwald, line 49: Brief comments defining the transformations

cell_classification/augmentation_pipeline.py
```diff
@@ -0,0 +1,80 @@
+import imgaug as ia
+import imgaug.augmenters as iaa
+from imgaug.augmentables.segmaps import SegmentationMapsOnImage
+from imgaug.augmentables.batches import Batch
+import numpy as np
+
+
+def augment_images(images, masks, augmentation_pipeline):
```
ngreenwald, line 8: We'll need to make sure the tf.record follows this same format of separate stacks for images and masks

cell_classification/augmentation_pipeline_test.py
```diff
@@ -0,0 +1,64 @@
...
+
+    # check if right types and shapes are returned
+    assert type(augmented_images) == np.ndarray
+    assert type(augmented_masks) == np.ndarray
+    assert augmented_images.dtype == np.float32
+    assert augmented_masks.dtype == np.int32
+    assert augmented_images.shape == images.shape
+    assert augmented_masks.shape == masks.shape
+
+    # check if images are augmented
```
ngreenwald, line 55: Perfect level of commenting here for the tests

ngreenwald (changes requested):
Looks great! Just a few areas where some more comments would be helpful

cell_classification/augmentation_pipeline.py
JLrumberger: This is an error. Should only be squeezed once. The problem is the following: `augmentation_pipeline.augment_batch_(batch)` adds a channel dimension to the mask if there's none, to ensure that input and output dim are the same, I think I should check if a channel dim got added, and if this happened it should be squeezed. Because as it is now, it would also squeeze the batch dim for batch size 1 which is unwanted.

cell_classification/augmentation_pipeline.py
JLrumberger: It doesn't change. I changed this to `shape=mask.shape` to make it more clear.

cell_classification/augmentation_pipeline.py
JLrumberger: Yep, this goes into the `prepare_example` function. We should do it anyways, because the actual mibi image will be the only thing that is stored as float, for the rest we'll use the minimal data type (i.e. uint8 for binary mask, uint16 instance_mask) to reduce i/o as much as possible.

cell_classification/augmentation_pipeline.py
ngreenwald: Won't the masks always have a channels dim?

We should loop over a couple different shapes for the testing data. Right now it's `[3, 100, 100, 2]`. We should probably check the case where both the first and/or last index is 1.

JLrumberger: Changed it so that it works for masks with or without channel dim and with different number of channels.

cell_classification/augmentation_pipeline.py
```diff
@@ -0,0 +1,92 @@
...
+        np.array:
+            The augmented masks
+    """
+    masks_ = [SegmentationMapsOnImage(mask, shape=mask.shape) for mask in masks]
+    batch = Batch(images=images, segmentation_maps=masks_)
+    batch = augmentation_pipeline.augment_batch_(batch)
+    augmented_masks = [mask.arr for mask in batch.segmentation_maps_aug]
+    augmented_masks = np.stack(augmented_masks, 0)
+
+    # remove additional channel from single channel masks
```
ngreenwald, line 30: Nice, this is more robust and easier to understand

cell_classification/augmentation_pipeline_test.py
```diff
@@ -0,0 +1,87 @@
...
+    # check if masks are still binary with the right labels
+    assert list(np.unique(augmented_masks)) == [0, 1, 2]
+
+    # check if images and masks where augmented with the same spatial augmentations approx.
+    assert np.abs(augmented_images[augmented_masks == 0].mean() - images[masks == 0].mean()) < 1
+    assert np.abs(augmented_images[augmented_masks == 1].mean() - images[masks == 1].mean()) < 1
+    assert np.abs(augmented_images[augmented_masks == 2].mean() - images[masks == 2].mean()) < 5
+
+    # check control flow for different number of samples per batch and channels
+    for b in [1, 2, 3]:
```
ngreenwald: This is great. Instead of a for loop inside the test, you can use pytest to automatically pass the correct variable. Then you can get rid of this separate section and test the batch sizes across the whole function
https://github.com/angelolab/toffy/blob/acd6df4d4b71484d2aae5629094cc5ebb7cd7bd6/toffy/tiling_utils_test.py#L405

```
parametrize = pytest.mark.parametrize

@parametrize('batch_num', [1, 2, 3])
@parametrize('chan_num', [1,2,3])
def test_augment_images(batch_num, chan_num)
...

masks = np.zeros([batch_num 100, 100, chan_num], dtype=np.int32)
```

cell_classification/augmentation_pipeline_test.py
```diff
@@ -0,0 +1,87 @@
...
+        for c in [1, 2, 3]:
+            images = np.zeros([b, 100, 100, c], dtype=np.float32)
+            masks = np.zeros([b, 100, 100, c], dtype=np.int32)
+            # if c == 1:
+            #     masks = np.squeeze(masks, axis=-1)
+            _, augmented_masks = augment_images(images, masks, augmentation_pipeline)
+            assert augmented_masks.shape == masks.shape
+
+    # check control flow for no channel dimensions for the masks
+    masks = np.zeros([b, 100, 100], dtype=np.int32)
```
ngreenwald: You can either keep this as a separate test case, or do something like this at the top:

`mask = np.zeros((batch_num, 100, 100)) if chan_num == 0 else mask = np.zeros((batch_num, 100, 100, chan_num)`

ngreenwald (changes requested):
Looks awesome, here's what I was talking about for letting pytest handle the for loop. There's often scenarios where you want to loop over different parameters to test, this is really useful.

ngreenwald:
Also, make sure to update your branch. Can either do it locally or from github. 
image

cell_classification/augmentation_pipeline_test.py
JLrumberger: Thanks for the suggestion, looks way better with the decorator!

cell_classification/augmentation_pipeline_test.py
JLrumberger: I'll just keep it as is, because it's easier to read I think.

PR #10, opened 2022-08-24
Added check_input fn and tests

JLrumberger:

Added the `check_input` function and according tests. Question is mainly if I do too much or too little checks and tests.

I added a fair number of exceptions to test if all input values make sense

**Remaining issues**

Could be that I did too much or too little checks and tests. Also could be that there is a more elegant way to do the tests.

### DIFF
--- a/cell_classification/segmentation_data_prep.py
+++ b/cell_classification/segmentation_data_prep.py
@@ -200,7 +200,8 @@ class SegmentationTFRecords:
         # make tfrecord path
         os.makedirs(self.tf_record_path, exist_ok=True)
 
-        # read conversion matrix
+        # CONVERSION MATRIX
+        # read the file
         self.conversion_matrix = pd.read_csv(self.conversion_matrix_path)
 
         # check if markers were selected or take all markers from conversion matrix
@@ -209,17 +210,43 @@ class SegmentationTFRecords:
         else:
             self.selected_markers = self.selected_markers
 
+            # check if selected markers are in conversion matrix
+            if not set(self.selected_markers).issubset(set(self.conversion_matrix.columns)):
+                raise ValueError("The selected markers are not in the conversion matrix")
+
+        # NORMALIZATION DICT
         # load or construct normalization dict
         if str(self.normalization_dict_path).endswith(".json"):
             self.normalization_dict = json.load(open(self.normalization_dict_path, "r"))
+
+            # check if selected markers are in normalization dict
+            if not set(self.selected_markers).issubset(set(self.normalization_dict.keys())):
+                raise ValueError("The selected markers are not in the normalization dict")
         else:
+            # function raises a generic FileNotFoundError if selected_marker file
+            # or segmentation_fname not in data_folders
             self.normalization_dict = self.calculate_normalization_matrix(
                 self.data_folders,
                 self.selected_markers,
             )
+        # check if normalization_quantile is in [0, 1]
+        if self.normalization_quantile < 0 or self.normalization_quantile > 1:
+            raise ValueError("The normalization_quantile is not in [0, 1]")
 
+        # CELL TYPE TABLE
         # load cell_types.csv
         self.cell_type_table = pd.read_csv(self.cell_table_path)
+
+        # check if cell_type_key is in cell_type_table
+        if self.cell_type_key not in self.cell_type_table.columns:
+            raise ValueError("The cell_type_key is not in the cell_type_table")
+        # check if segment_label_key is in cell_type_table
+        if self.segment_label_key not in self.cell_type_table.columns:
+            raise ValueError("The segment_label_key is not in the cell_type_table")
+
+        # check if sample_key is in cell_type_table
+        if self.sample_key not in self.cell_type_table.columns:
+            raise ValueError("The sample_key is not in the cell_type_table")
 
     def make_tf_record(self, data_folders):
         """Iterates through the data_folders and loads, transforms and
--- a/cell_classification/segmentation_data_prep.py
+++ b/cell_classification/segmentation_data_prep.py
@@ -270,13 +270,11 @@ class SegmentationTFRecords:
         if self.sample_key not in self.cell_type_table.columns:
             raise ValueError("The sample_key is not in the cell_type_table")
 
-        # check if sample_names in cell_type_table match sample_names in data_folders
-        for sample_name in self.cell_type_table[self.sample_key].values:
-            matches = [folder for folder in self.data_folders if re.search(sample_name, folder)]
-            if not matches:
-                raise ValueError(
-                    "The sample_name {} is not in the data_folders".format(sample_name)
-                )
+        # check if sample_names in cell_type_table match sample_names in data_folder
+        verify_in_list(
+            sample_names=self.cell_type_table[self.sample_key].values,
+            data_folders=list_folders(self.data_dir)
+        )
 
     def make_tf_record(self, data_folders):
         """Iterates through the data_folders and loads, transforms and
@@ -286,7 +284,6 @@ class SegmentationTFRecords:
             tf_record_path (str):
                 The path to the tf record to make
         """
-
         return None
 
     def calculate_normalization_matrix(self, data_folders, selected_markers):
--- a/cell_classification/segmentation_data_prep.py
+++ b/cell_classification/segmentation_data_prep.py
@@ -8,24 +8,25 @@ import os
 import xarray
 from tqdm import tqdm
 import json
+from ark.utils.misc_utils import verify_in_list
+from ark.utils.io_utils import list_folders, validate_paths
+import re
 
 
 class SegmentationTFRecords:
     """Prepares the data for the segmentation model"""
 
     def __init__(
-        self, data_folders, cell_table_path, conversion_matrix_path,
-        imaging_platform, dataset, tile_size, tf_record_path,
-        selected_markers=None, normalization_dict_path=None,
-        normalization_quantile=0.99, cell_type_key="cluster_labels",
-        sample_key="SampleID", segmentation_fname="cell_segmentation",
-        segment_label_key="labels",
+        self, data_dir, cell_table_path, conversion_matrix_path, imaging_platform, dataset,
+        tile_size, tf_record_path, selected_markers=None, normalization_dict_path=None,
+        normalization_quantile=0.99, cell_type_key="cluster_labels", sample_key="SampleID",
+        segmentation_fname="cell_segmentation", segment_label_key="labels",
     ):
         """Initializes SegmentationTFRecords and loads everything except the images
 
         Args:
-            data_folders (list):
-                List of folders containing the multiplexed imaging data
+            data_dir str:
+                Path where the data is stored
             cell_table_path (str):
                 Path to the cell table
             conversion_matrix_path (str):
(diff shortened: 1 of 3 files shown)


cell_classification/segmentation_data_prep.py
```diff
@@ -209,18 +210,44 @@ def check_input(self):
         else:
             self.selected_markers = self.selected_markers
 
+            # check if selected markers are in conversion matrix
+            if not set(self.selected_markers).issubset(set(self.conversion_matrix.columns)):
```
ngreenwald: We created helper functions for checking the different ways certain arguments relate to others (must be a subset, must be equal, must be equal and in the same order, etc): https://github.com/angelolab/toffy/blob/main/toffy/rosetta.py#L83. This function gives some good examples of how to use them. In addition, because these are user-facing error checks, in addition to reporting the error, it also will find and print out the specific list items that are different. 

Right now `ark` has a lot of requirements, so long-term I don't think we want to import it. However, we're going to be moving all of these helper functions to their own repo next month, so for now I think it's okay.

cell_classification/segmentation_data_prep.py
```diff
@@ -209,18 +210,44 @@ def check_input(self):
         else:
             self.selected_markers = self.selected_markers
 
+            # check if selected markers are in conversion matrix
```
ngreenwald: I think `data_folders` can get changed to `data_dir`. We should add a check that this is a valid directory. Then `data_folders` can be generated automatically with `ark.utils.io_utils.list_folders` by listing all of the folders within the directory. We'll have hundreds to thousands of folders in these datasets

cell_classification/segmentation_data_prep.py
```diff
@@ -209,18 +210,44 @@ def check_input(self):
...
+            # or segmentation_fname not in data_folders
             self.normalization_dict = self.calculate_normalization_matrix(
                 self.data_folders,
                 self.selected_markers,
             )
+        # check if normalization_quantile is in [0, 1]
+        if self.normalization_quantile < 0 or self.normalization_quantile > 1:
+            raise ValueError("The normalization_quantile is not in [0, 1]")
 
+        # CELL TYPE TABLE
```
ngreenwald, line 257: We should check that this path is valid with [validate_path](https://github.com/angelolab/ark-analysis/blob/main/ark/utils/io_utils.py#L7)

cell_classification/segmentation_data_prep.py
```diff
@@ -209,18 +210,44 @@ def check_input(self):
...
 
+        # check if cell_type_key is in cell_type_table
+        if self.cell_type_key not in self.cell_type_table.columns:
+            raise ValueError("The cell_type_key is not in the cell_type_table")
+        # check if segment_label_key is in cell_type_table
+        if self.segment_label_key not in self.cell_type_table.columns:
+            raise ValueError("The segment_label_key is not in the cell_type_table")
+
+        # check if sample_key is in cell_type_table
+        if self.sample_key not in self.cell_type_table.columns:
```
ngreenwald, line 270: We should also make sure that all of the unique cell types in the cell table are present in the conversion matrix

cell_classification/segmentation_data_prep.py
```diff
@@ -209,18 +210,44 @@ def check_input(self):
...
+            raise ValueError("The sample_key is not in the cell_type_table")
```
ngreenwald, line 271: We should check that all of the sample names in the cell table match the data_folder names

cell_classification/segmentation_data_prep_test.py
```diff
@@ -122,15 +123,34 @@ def test_check_input():
...
-        data_folders = prepare_test_data_folders(5, temp_dir, norm_dict.keys())
+        with open(os.path.join(temp_dir, "norm_dict.json"), "w") as f:
+            json.dump(norm_dict, f)
+        data_folders = prepare_test_data_folders(5, temp_dir, list(norm_dict.keys()) + ["XYZ"])
         cell_table_path = os.path.join(temp_dir, "cell_type_table.csv")
         cell_table = prepare_cell_type_table()
         cell_table.to_csv(cell_table_path, index=False)
 
+        # CONVERSION MATRIX
+        # check if conversion_matrix is loaded correctly in check_input
```
ngreenwald, line 137: The purpose of the check_input function is to find any errors in the parameters the user provided. Since we're not relying on this function to append the conversion matrix as attribute or anything, I don't think it's necessary to have a correctness check here.

cell_classification/segmentation_data_prep_test.py
```diff
@@ -122,15 +123,34 @@ def test_check_input():
...
+            cell_table_path=cell_table_path,
+            normalization_dict_path=os.path.join(temp_dir, "norm_dict.json"),
+        )
+        data_prep.check_input()
+        assert np.array_equal(data_prep.conversion_matrix, conversion_matrix)
+        data_prep_working = copy.deepcopy(data_prep)
+
+        # check if ValueError is raised when selected_markers not in conversion_matrix
+        data_prep.selected_markers = ["XYZ"]
+        with pytest.raises(ValueError):
```
ngreenwald: These are the types of checks we definitely want in the testing for this function

cell_classification/segmentation_data_prep_test.py
```diff
@@ -142,15 +162,68 @@ def test_check_input():
...
+            columns=["CD11c", "CD14", "CD56", "CD57", "XYZ", "ZYX"],
+            index=["stromal", "FAP", "NK", "CD4T", "CD14", "CD163"],
+        )
+        conversion_matrix_path = os.path.join(temp_dir, "conversion_matrix.csv")
+        conversion_matrix.to_csv(conversion_matrix_path, index=False)
+        data_prep = copy.deepcopy(data_prep_working)
+        data_prep.conversion_matrix_path = conversion_matrix_path
+        data_prep.normalization_dict_path = os.path.join(temp_dir, "norm_dict.json")
+        data_prep.selected_markers = ["ZYX"]
+        with pytest.raises(ValueError):
```
ngreenwald: When there is a function that can raise lots of different potential errors, it's a good idea to check not only that right type of error is being raised, but that the exact correct error was raised. You can use the `match` substring to do this
https://github.com/angelolab/ark-analysis/blob/main/ark/utils/io_utils_test.py#L48

cell_classification/segmentation_data_prep_test.py
```diff
@@ -142,15 +162,68 @@ def test_check_input():
...
-            conversion_matrix_path=conversion_matrix_path,
-            tf_record_path=os.path.join(temp_dir, "tf_record_path"),
-            cell_table_path=cell_table_path,
-        )
+        data_prep.data_folders = data_folders
+        data_prep.normalization_dict_path = None
         data_prep.check_input()
         assert norm_dict == data_prep.normalization_dict
 
+        # check if ValueError is raised if selected_markers in conversion_matrix
```
ngreenwald, line 174: We should add a direct check to make sure that selected_markers is present in one of the example images. 

If the user says "CD455" but means CD45, makes more sense to raise an error directly saying "CD455" not found in image data, rather than saying CD45 not found in conversion matrix

cell_classification/segmentation_data_prep_test.py
```diff
@@ -142,15 +162,68 @@ def test_check_input():
...
+        data_prep = copy.deepcopy(data_prep_working)
+        data_prep.conversion_matrix_path = conversion_matrix_path
+        data_prep.normalization_dict_path = os.path.join(temp_dir, "norm_dict.json")
+        data_prep.selected_markers = ["ZYX"]
+        with pytest.raises(ValueError):
+            data_prep.check_input()
+
+        # check if FileNotFoundError is raised if data_folders and conversion_matrix_path are given
+        # together with selected_markers were images are missing for in data_folders
+        data_prep = copy.deepcopy(data_prep_working)
```
ngreenwald, line 199: This is a very indirect way of finding out that the image data does not match up with the specified markers. Better to find out earlier. 

In addition, we generally only want to check the behavior of the function directly under test. This checks the behavior of a function that check_input is calling and the fucntion that function is calling etc. 

Just get rid of this one

cell_classification/segmentation_data_prep_test.py
```diff
@@ -142,15 +162,68 @@ def test_check_input():
...
+        with pytest.raises(ValueError):
+            data_prep.check_input()
+
+        # CELL TYPE TABLE
+        # check if cell_type_table is loaded correctly in check_input
+        # when cell_type_table_path is given to init
+        data_prep = copy.deepcopy(data_prep_working)
+        data_prep.cell_type_table_path = cell_table_path
+        data_prep.check_input()
+        assert np.array_equal(cell_table, data_prep.cell_type_table)
```
ngreenwald, line 219: Same here, don't need to check this behavior, only the error checking functionality. 

Unless this is the only place where the cell table is loaded, then makes sense to check. But for files that are only loaded in this function in order for those files to be checked, and are loaded elswhere, then it's okay to skip.

ngreenwald (changes requested):
Looks great!

cell_classification/segmentation_data_prep.py
JLrumberger: So in the above three comments you write:
1. supply data_dir instead of data_folders, get the folders within `check_input`
2. check if unique cell types from cell table are present in the conversion matrix
3. check that all of the sample names in the cell table match data_folder names

I need to think about this more, but I think these could be overly restrictive. If we have a dataset where some FOVs should be excluded, 1. would be too restrictive and 3. would throw an exception. Also 2. could be better checked when it's needed in `get_marker_activity` and not here, since it could be that samples are excluded that have a certain cell type that is not in the conversion matrix. I worked mostly with super messy data and I just see that the more assumptions and the less degree of freedoms we'll have here, the more we need to change this stuff later.

ngreenwald: Agreed, lets skip these for now.

cell_classification/segmentation_data_prep_test.py
JLrumberger: I renamed the function to `load_and_check_input` since outside of the images, all inputs are loaded in this function so I also keep the checks there.

cell_classification/segmentation_data_prep.py
```diff
@@ -195,32 +196,88 @@ def tile_example(example, tile_size):
...
+
+        # check if sample_names in cell_type_table match sample_names in data_folders
+        for sample_name in self.cell_type_table[self.sample_key].values:
+            matches = [folder for folder in self.data_folders if re.search(sample_name, folder)]
```
ngreenwald: This can be `verify_in_list` as well

cell_classification/segmentation_data_prep_test.py
```diff
@@ -115,44 +124,137 @@ def test_calculate_normalization_matrix():
...
+        )
+        data_prep.load_and_check_input()
+        assert np.array_equal(data_prep.conversion_matrix, conversion_matrix)
+        data_prep_working = copy.deepcopy(data_prep)
+
+        # check if ValueError is raised when selected_markers not in conversion_matrix
+        data_prep.selected_markers = ["XYZ"]
+        with pytest.raises(
+            ValueError,
+            match="Not all values given in list selected markers were found in list conversion "
```
ngreenwald: You can also use a partial match, as long as it's unique to that error

ngreenwald (changes requested):
Looks good. You should also add coveralls to the repo (can ask Sri for help after the standup today)

ngreenwald (approved):

Issue #6, opened 2022-08-19
Overview of tasks for model pipeline


ngreenwald:
The following is a summary of potential improvements/features for the classification model

[Experiment Plan](https://docs.google.com/spreadsheets/d/1bj_t5tNHluGJoujQYqk08CEP_8t5uI1prBFmbS3YyNc/edit?usp=sharing)

Data normalization:
- [x] Histogram normalization across the whole dataset. **MVP implementation for this task.** The simplest way to normalize the data is to calculate the distribution of each marker across the entire training dataset, then normalize the data. 
- [x] Image-specific normalization. Investigate if there is a way to normalize each image without needing to know the distribution of the whole dataset.

Dataset creation:
- [x] [MIBI Segmentation Data Loader](https://github.com/angelolab/cell_classification/issues/1). **MPV implementation for this task**. This would create a segmentation based scheme for data loading
- [x] Plot utils to visualize positive/negative/unspecific (1/0/-1) cells on top of a marker channel
- [ ] Multi-channel segmentation data loader. This would allow the model to be trained on multiple channels and labels simultaneously
- [x] Fixed channels to aid prediction. This would allow the user to specify a list of channels that would always be included in a fixed order as model inputs, such as membrane/nucleus channels, to give cell morphology or co-expression information
- [x] Remove ambiguous cells from training data. There are certain cells that are likely borderline expression for specific markers. We should consider adding an option to remove them from the training data completely. Obviously removing hard examples will by definition improve performance. So we should evaluate if this improves performance on the rest of the dataset independent of removing these tricky ones. 
- [ ] Normalize data once and story it on disc/do it in-place to avoid the compute overhead from normalizing it each time we generate a new tfrecord dataset.

Image augmentation
- [x] Basic image augmentation for segmentation model. **MVP implementation for this task.** This would augment the input image channel separate from the input masks
- [x] Multi-channel augmentation. This would handle multiple input channels of image data. 
- [x] Use augmentations on the GPU to speed up training (e.g. via [kerasCV ](https://keras.io/guides/keras_cv/cut_mix_mix_up_and_rand_augment/) or [tensorlayer](https://tensorlayer.readthedocs.io/en/v2.2.4/modules/prepro.html?highlight=elastic))
- [ ] Investigate effect of augmentation on val loss across dataset sizes, in line with [this augmentation paper](https://arxiv.org/abs/2210.06441)
- [ ] Determine if MixUp provides an additional performance boost over current augmentations

Image generator
[...]

PR #14, opened 2022-09-02
Prepare example

### LINKED ISSUE #1: Mibi Segmentation Dataloader
**Section 1: Design details**

**Relevant background**

To train and validate an ML model that predicts if a given marker is positive or negative for a cell, we need a tfrecords object that stores and returns input and ground truth data as well as meta data (dataset, platform, etc.) of the samples. 

**Design overview**

We'll be using tfrecords for storage and fast i/o during training. For constructing the tfrecords object, we need to construct for each single training example a dictionary that holds all necessary data. For this we develop a number of helper functions that load data from different sources (mibi tifs, dfs holding cell type information, etc.). 

We want our training examples to consist of the following data
- single or multiple channels from multiplexed images (MIBI, CODEX, etc.)
- name of each channel
- name of the dataset
- the imaging platform used to acquire the images
- the cell types as semantic segmentation or as a dictionary
- cell segmentation masks in binary form (either 0/1 => background&cell boundary/cell interior or 0/1/2 => background/cell interior/cell boundary
- ground truth binary segmentation labels for each of the channels

**Design list/flowchart**

The desired output is a tfrecords store that stores all of the above mentioned data. Therefore we need:

1. prepare example class that takes in all required information (list of folder names, path to conversion matrix and cell table, etc.) and serializes the training examples into a tfrecord store. The function will make use of the following helper functions:
2. function that takes an instance segmentation mask and returns a binary mask
3. function that takes a list of cell segmentation indices and returns a list of cell types
4. function that takes a list of cell types and a marker name and returns marker activity as a list that contains 
    - "1" if the marker is positive for the cell type
    - "0" if the marker is negative for the cell type
    - "-1" if the marker is unspecific for the cell type (e.g Vimentin for endothelial cells)
5. function that takes in instance segmentation mask and marker activity list and returns marker activity ground truth mask
6. helper function that iterates through the whole dataset to calculate 99% quantile for each channel for the whole dataset and stores the results in a .json and returns it as a dict

**Required inputs**

- list of folder paths of multiplexed images (height x width) stored as .tiff
- path to conversion matrix (cell types x markers) that holds for each cell type which marker is positive/negative/unspecific
- path to `cell_table.csv` (cells x features) that holds the cell types for each cell
- imaging platform (str)
- dataset (str)
- tile-size [int,int] optional
- key (str) for the cell type column in `cell_table.csv`, default `cluster_labels`
- key (str) for cell segmentation tif, default `cell_segmentation` 

**Output files**

- tfrecord store holding all training samples, consisting of
    - multiplexed image: H x W x C
    - instance segmentation (binary erosion) mask: H x W x 1
    - ground truth mask: H x W x C
    - either instance segmentation mask: H x W x 1 and list of cell types (#cells in ROI) or semantic segmentation mask H x W x 1
    - name of each channel
    - name of the dataset
    - name of the imaging platform
- .json file holding the 99%-quantile for each channel for the whole dataset

Note to self: store data as .jpeg encoded byte arrays in the tf_record as described in this issue [tf issue 9675](https://github.com/tensorflow/tensorflow/issues/9675).

**Control flow**

Provide additional, more granular details (if necessary) about how the proposed coding logic will be put together

**Milestones and timeline**

List each of the major components of the project, and provide an estimated completion date for each one.

JLrumberger:

Adding the remaining functionalities `make_tf_record`, `serialize_example`, `tile_example` and `prepare_example` as well as the parse function `parse_dict` that takes a tfrecord example consisting of byte strings and parses it into a dict of tensors, dataframes and strings. This PR closes #1.

Added the above mentioned functionality to `segmentation_data_prep.py` and wrote the tests for it.

**Remaining issues**

None hopefully. I could add a `main` method and an `argparse` CLI if this is desired?

### DIFF
--- a/cell_classification/segmentation_data_prep.py
+++ b/cell_classification/segmentation_data_prep.py
@@ -15,7 +15,7 @@ class SegmentationTFRecords:
 
     def __init__(
         self, data_folders, cell_table_path, conversion_matrix_path,
-        imaging_platform, dataset, tile_size, tf_record_path, selected_markers=None,
+        imaging_platform, dataset, tile_size, stride, tf_record_path, selected_markers=None,
         normalization_dict_path=None, normalization_quantile=0.99,
         cell_type_key="cluster_labels", sample_key="SampleID",
         segmentation_fname="cell_segmentation", segment_label_key="labels",
@@ -35,6 +35,8 @@ class SegmentationTFRecords:
                 The dataset where the imaging data comes from
             tile_size list [int,int]:
                 The size of the tiles to use for the segmentation model
+            stride list [int,int]:
+                The stride to tile the data
             tf_record_path (str):
                 The path to the tf record to make
             selected_markers (list):
@@ -66,6 +68,8 @@ class SegmentationTFRecords:
         self.tf_record_path = tf_record_path
         self.cell_type_key = cell_type_key
         self.cell_table_path = cell_table_path
+        self.tile_size = tile_size
+        self.stride = stride
 
     def get_image(self, data_folder, marker):
         """Loads the images from a single data_folder
@@ -188,7 +192,7 @@ class SegmentationTFRecords:
             "cell_types": cell_types,
         }
 
-    def tile_example(self, example, tile_size, stride, spatial_keys=[
+    def tile_example(self, example, spatial_keys=[
             "mplex_img", "binary_mask", "instance_mask", "marker_activity_mask",
         ],
     ):
@@ -212,8 +216,8 @@ class SegmentationTFRecords:
             if example[key].ndim == 2:
                 example[key] = np.expand_dims(example[key], axis=-1)
             res = np.lib.stride_tricks.sliding_window_view(
-                example[key], window_shape=tile_size + list(example[key].shape[2:])
-            )[:: stride[0], :: stride[1]]
+                example[key], window_shape=self.tile_size + list(example[key].shape[2:])
+            )[:: self.stride[0], :: self.stride[1]]
             sh = list(res.shape)
             tiled_examples[key] = res.reshape([sh[0] * sh[1]] + sh[3:])  # tiles x H x W x C
 
--- a/cell_classification/segmentation_data_prep.py
+++ b/cell_classification/segmentation_data_prep.py
@@ -16,10 +16,9 @@ class SegmentationTFRecords:
     """Prepares the data for the segmentation model"""
 
     def __init__(
-        self, data_dir, cell_table_path, conversion_matrix_path,
-        imaging_platform, dataset, tile_size, stride, tf_record_path, selected_markers=None,
-        normalization_dict_path=None, normalization_quantile=0.99,
-        cell_type_key="cluster_labels", sample_key="SampleID",
+        self, data_dir, cell_table_path, conversion_matrix_path, imaging_platform, dataset,
+        tile_size, stride, tf_record_path, selected_markers=None, normalization_dict_path=None,
+        normalization_quantile=0.99, cell_type_key="cluster_labels", sample_key="SampleID",
         segmentation_fname="cell_segmentation", segment_label_key="labels",
     ):
         """Initializes SegmentationTFRecords and loads everything except the images
@@ -105,9 +104,7 @@ class SegmentationTFRecords:
             np.array:
                 The instance mask
         """
-        instance_mask = imread(
-            os.path.join(data_folder, self.segmentation_fname + ".tiff")
-        )
+        instance_mask = imread(os.path.join(data_folder, self.segmentation_fname + ".tiff"))
         if instance_mask.ndim == 2:
             instance_mask = np.expand_dims(instance_mask, axis=-1)
         edge = find_boundaries(instance_mask, mode="inner").astype(np.uint8)
@@ -128,9 +125,7 @@ class SegmentationTFRecords:
                 The marker activity for the given labels, 1 if the marker is active, 0
                 otherwise and -1 if the marker is not specific enough to be considered active
         """
-        sample_subset = self.cell_type_table[
-            self.cell_type_table.SampleID == sample_name
-        ]
+        sample_subset = self.cell_type_table[self.cell_type_table.SampleID == sample_name]
         cell_types = sample_subset[self.cell_type_key].values
 
         df = pd.DataFrame(
@@ -179,13 +174,9 @@ class SegmentationTFRecords:
         mplex_img = self.get_image(data_folder, marker)
         mplex_img /= self.normalization_dict[marker]
         binary_mask, instance_mask = self.get_inst_binary_masks(data_folder)
-
+        fov = os.path.split(data_folder)[-1]
         # get the cell types and marker activity mask
-        marker_activity, cell_types = self.get_marker_activity(
-            os.path.split(data_folder)[-2],
-            self.conversion_matrix,
-            marker
-        )
+        marker_activity, cell_types = self.get_marker_activity(fov, self.conversion_matrix, marker)
         marker_activity_mask = self.get_marker_activity_mask(
             instance_mask, binary_mask, marker_activity
         )
@@ -198,9 +189,11 @@ class SegmentationTFRecords:
             "dataset": self.dataset,
             "marker": marker,
             "cell_types": marker_activity,
+            "folder_name": fov,
         }
 
-    def tile_example(self, example, spatial_keys=[
+    def tile_example(
+        self, example, spatial_keys=[
             "mplex_img", "binary_mask", "instance_mask", "marker_activity_mask",
         ],
     ):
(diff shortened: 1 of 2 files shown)


cell_classification/segmentation_data_prep.py
```diff
@@ -167,12 +174,11 @@ def prepare_example(self, data_folder, marker):
         mplex_img = self.get_image(data_folder, marker)
         mplex_img /= self.normalization_dict[marker]
```
ngreenwald, line 177: I think eventually it will make more sense to have this normalization happen in a different function. Probably we'll want to avoid re-normalizing the data every time we load it in, and instead normalizing once and then loading the normalized data each time would make more sense. Not necessary for the first pass though, so if you want to just make an issue about it and we can fix it later once it becomes a bottleneck that's totally fine

cell_classification/segmentation_data_prep.py
```diff
@@ -182,19 +188,53 @@ def prepare_example(self, data_folder, marker):
...
+        ],
+    ):
         """Tiles the example into a grid of tiles
         Args:
             example (dict):
                 The example to tile
+            tile_size (list):
+                The size of the tiles
+            stride (list):
+                The stride of the tiles
```
ngreenwald: These are attributes now, not args

cell_classification/segmentation_data_prep.py
```diff
@@ -182,19 +188,53 @@ def prepare_example(self, data_folder, marker):
...
+            if example[key].ndim == 2:
+                example[key] = np.expand_dims(example[key], axis=-1)
+            res = np.lib.stride_tricks.sliding_window_view(
+                example[key], window_shape=self.tile_size + list(example[key].shape[2:])
+            )[:: self.stride[0], :: self.stride[1]]
+            sh = list(res.shape)
+            tiled_examples[key] = res.reshape([sh[0] * sh[1]] + sh[3:])  # tiles x H x W x C
+
+        # store individual tiled examples in a list of dicts
+        non_spatial_keys = [key for key in example if key not in spatial_keys]
```
ngreenwald: This is fine for now. We should add an issue though that the non-spatial metadata for each tile will not have been subsetted to include only cells in the tile, but all cells from the original untiled image. If we decide to use these non-spatial keys for assessing accuracy, we'll need to either subset then, or decide to change this part and do subsetting here, depending on which makes more sense

cell_classification/segmentation_data_prep.py
```diff
@@ -273,19 +317,81 @@ def load_and_check_input(self):
...
+                    example_list = self.tile_example(example)
+                else:
+                    example_list = [example]
+
+                # serialize and write examples to tfrecord
+                for example in example_list:
+                    example_serialized = self.serialize_example(example)
+                    self.writer.write(example_serialized)
+        self.writer.close()
         return None
```
ngreenwald: Don't need the empty return anymore

cell_classification/segmentation_data_prep_test.py
```diff
@@ -334,5 +367,115 @@ def test_get_marker_activity_mask():
...
+        "cell_types": ["CD11c", "CD11b", "CD11a"],
+        "marker": "CD11c",
+    }
+    data_prep = prep_object(tile_size=[128, 128], stride=[128, 128])
+    tiled_examples = data_prep.tile_example(example)
+
+    # check if the correct number of tiles got returned
+    assert len(tiled_examples) == 16
+
+    # check if the correct spatial dimensions got returned and dtype is correct
```
ngreenwald, line 403: Can you also add a check that the first tiled example is equal to the top left corner of the original untiled image, last one is bottom right, or something along those lines?

ngreenwald (changes requested):
This looks great, really minor changes only. 

1. Since this is the last PR for the data loader design doc #1, you can edit the description of this PR to close that issue.
2. You can also get started on a design doc for the next part of the pipeline that you want to work on
3. You should get Sri or Alex to help you install coveralls in this repo so that we can see what parts of the code aren't being tested
4. Although the tests are passing, it could be that there are still some bugs hiding in here. I think it will be useful to write a couple very simple functions to take a serialized dataset, load an example, and plot the the raw data, masks, and labels to make sure everything is working. This will also probably be useful for inspecting the model output further down the road.

cell_classification/segmentation_data_prep.py
JLrumberger: Normalizing is really cheap once the `normalization_dict` is calculated and we don't need to have a copy of the dataset stored on disc. I put it into the backlog #6.

cell_classification/segmentation_data_prep.py
JLrumberger: That's a good idea. I just implemented it since I don't really see a downside. If we see a problem with this in the feature I'll make it optional.

cell_classification/segmentation_data_prep_test.py
JLrumberger: Yep, I added tests for both cases for all spatial keys.

JLrumberger:
> This looks great, really minor changes only.
> 
> 1. Since this is the last PR for the data loader design doc #1, you can edit the description of this PR to close that issue.
> 2. You can also get started on a design doc for the next part of the pipeline that you want to work on
> 3. You should get Sri or Alex to help you install coveralls in this repo so that we can see what parts of the code aren't being tested
> 4. Although the tests are passing, it could be that there are still some bugs hiding in here. I think it will be useful to write a couple very simple functions to take a serialized dataset, load an example, and plot the the raw data, masks, and labels to make sure everything is working. This will also probably be useful for inspecting the model output further down the road.

1. Done
2. Will do
3. Done I think, at least I see coveralls in the travis checks
4. I'd do it with the real data in a notebook, to check visually if everything makes sense. Or is there a better way to do this?

cell_classification/segmentation_data_prep.py
```diff
@@ -182,19 +189,57 @@ def prepare_example(self, data_folder, marker):
...
+        for tile in range(num_tiles):
+            example_out = {}
+            for key in spatial_keys:
+                example_out[key] = tiled_examples[key][tile]
+            for non_spatial_key in non_spatial_keys:
+                example_out[non_spatial_key] = example[non_spatial_key]
+
+            # subset marker_activity to the labels that are present in the tile
+            label_subset = np.unique(example_out["instance_mask"]).astype(np.uint16).tolist()
+            example_out["marker_activity"] = example["marker_activity"].loc[
```
ngreenwald: Since `example_out` is a dictionary with each key being a different data structure, having good names is important for making the code easy to read. I thought initially that `marker_activity` was the actual list, but it's actually a `df`, where the column `activity` within the df has the activities. However, to figure that out I had to go look at the test data you constructed. 

Most of the other keys in the example input have very clear names, but changing this to `activity_df` or something similar would be helpful

cell_classification/segmentation_data_prep.py
```diff
@@ -273,18 +322,79 @@ def load_and_check_input(self):
...
+
+        # iterate through data_folders and markers to prepare and tile examples
+        print("Preparing examples...")
+        for data_folder in tqdm(self.data_folders):
+            for marker in self.selected_markers:
+                example = self.prepare_example(data_folder, marker)
+                if self.tile_size:
+                    example_list = self.tile_example(example)
+                else:
+                    example_list = [example]
```
ngreenwald, line 366: This is what coveralls is really good for: identifying if there are parts of your control flow that aren't being tested. Any line that never gets run by the tests shows up in red:
image

Adding a simple test case without tiling will be good to make sure this logic works as expected. It would also probably be worthwhile to see what happens when the tile size doesn't divide evenly into the image size, when they're exactly equal, etc.

cell_classification/segmentation_data_prep_test.py
```diff
@@ -334,5 +367,145 @@ def test_get_marker_activity_mask():
...
+        "mplex_img": np.random.rand(512, 512, 3).astype(np.float32),
+        "binary_mask": np.random.randint(0, 2, [512, 512, 1]).astype(np.uint8),
+        "instance_mask": instance_mask,
+        "marker_activity_mask": np.random.randint(0, 2, [512, 512, 21]).astype(np.uint8),
+        "dataset": "test_dataset",
+        "platform": "mibi",
+        "marker_activity": marker_activity,
+        "marker": "CD11c",
+    }
+    data_prep = prep_object(tile_size=[128, 128], stride=[128, 128])
```
ngreenwald: This can be pulled out into a call to `@parametrize` to make the tests more robust. I can imagine there will be cases where the dataset is exactly divisible by our tile, where there is a remainder, where there's only one tile, etc etc

ngreenwald (changes requested):
Looks great! Just a minor comment about test coverage and some naming. 

For inspecting the output, I agree, using the real data definitely makes the most sense. I can give you the updated dataset tomorrow.

cell_classification/segmentation_data_prep.py
JLrumberger: Wow, looks like a great tool! Thanks for the explanation. I added default zero padding for the case that images are not divisible by the tile_size because I think this is always desired. If my assumption turns out to be wrong, I'll make it optional at a later stage.

cell_classification/segmentation_data_prep_test.py
JLrumberger: Done! This actually uncovered some small mistakes here and there in the code and tests and took way longer than I expected but ultimately improved code quality :>

cell_classification/segmentation_data_prep.py
ngreenwald: Agreed

cell_classification/segmentation_data_prep_test.py
ngreenwald: Awesome, that's the goal! Find them now

ngreenwald (approved):

PR #16, opened 2022-09-07
Plot utils

### LINKED ISSUE #15: Plot utils
**Section 1: Design details**

To check if dataset creation with #1 gives correct results on real world data, we'll use a plot helper function that takes in a marker channel and the positive/negative/unspecific (1/0/2) channel and visualizes the cell boundaries in different colors for the different classes. It should also have the option to write PNGs to path if available.

**Relevant background**

We want to generate plots for sanity checking the generated labels and input data which is produced by #1. 

**Design overview**

We will load a serialized example produced by #1, use the `parse_example` function to make it a dictionary. The dictionary is then passed to a plotting function that plots the marker image and transforms the marker activity mask to boundaries. Then the boundaries are colored based on their value (1/0/2) and overlayed on the plot. A legend tells you the color of the marker values in the overlay.

**Design list/flowchart**

1. dataloader: loads examples from tfrecord
2. parse function: parsed loaded examples to dict
3. plot function: takes dict, calls boundary transform plotting function and returns plot or saves it to path
4. boundary transform plotting helper function: takes in a mask image, generates boundaries and assigns values to the boundaries based on the value of the object.

**Required inputs**

- serialized tfrecord examples with keys `["mplex_img", "binary_mask", "instance_mask", "imaging_platform", "marker_activity_mask", "dataset", "marker", "activity_df", "folder_name"]`
- path to store plots

**Output files**

Plots showing an overlay of marker channels and marker activity.

Once you have completed section 1, please tag the relevant parties and iterate on the initial design details until everyone is satisfied. Then, 
proceed to section 2

**Control flow**

Provide additional, more granular details (if necessary) about how the proposed coding logic will be put together

**Milestones and timeline**

List each of the major components of the project, and provide an estimated completion date for each one.

JLrumberger:
**If you haven't already, please read through our [contributing guidelines](https://ark-analysis.readthedocs.io/en/latest/_rtd/contributing.html) before opening your PR**

This PR adds some plotting utils to inspect the examples in the tfrecord to close #15. Also it adds small changes to `segmentation_data_prep.py` based on errors that occured when wrangling real data with it.

There are two plotting functions:
- `plot_overlay` which plots the boundaries of all cells overlayed onto the marker image with positive cells highlighted in green, unspecific cells highlighted in blue and the rest highlighted in gray.
- `plot_together` which plots the marker image, the binary cell masks and the ground truth marker activity mask (again positive/unspecific are highlighted in green/blue).

Another key change is that `SegmentationTFRecords` now uses the 0.999 quantile by default for normalization of the marker image and clips all values above it. In addition, the class also sets `cell_table[cell_type_key]` and `conversion_matrix.index` to lower case. This could have negative consequences if there exists a cell type name that should be case sensitive. 

**Remaining issues**

Test coverage of `plot_utils.py` is fairly low, because I didn't write tests for the CLI.

### DIFF
--- a/cell_classification/plot_utils.py
+++ b/cell_classification/plot_utils.py
@@ -9,13 +9,14 @@ import numpy as np
 import argparse
 from tqdm import tqdm
 
+
 def segmentation_to_boundaries(labels):
-    """ Convert a segmentation to a binary mask of the boundaries
-    
+    """Convert a segmentation to a binary mask of the boundaries
+
     Args:
         segmentation (np.ndarray):
             A 2D array of integers representing a segmentation
-    
+
     Returns:
         np.ndarray:
             A 2D array of booleans representing the boundaries of the segmentation
@@ -25,9 +26,9 @@ def segmentation_to_boundaries(labels):
     return boundaries
 
 
-def plot_overlay(example, dpi=None, save_dir=None, save_file=None):
-    """ Plot the marker image and the marker activity segmentation overlayed
-    
+def plot_overlay(example, save_dir=None, save_file=None, dpi=160):
+    """Plot the marker image and the marker activity segmentation overlayed
+
     Args:
         example (dict):
             Dictionary with keys "mplex_img", "marker_activity_mask"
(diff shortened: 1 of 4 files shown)


ngreenwald (approved):
Looks good! The plotting stuff will probably evolve a bit once we start looking at model outputs, don't think it's worth spending too much time cleaning it up now (if ever)

Issue #17, opened 2022-09-12
Put argparse CLI in plot_utils in a separate function and write tests for it


JLrumberger:
**Is your feature request related to a problem? Please describe.**

Test coverage in `plot_utils` is too low, to increase it we need to also test the CLI. Thus we need to put the CLI related code from main into it's own function and then write tests for it

**Describe the solution you'd like**

Put all the argparse stuff from main into its own function

**Describe alternatives you've considered**

None

ngreenwald:
This will be good to do eventually, but definitely not top priority.

PR #18, opened 2022-09-12
Training script

JLrumberger:

This PR adds the training script for the cell classification model. I didn't open an issue beforehand so this PR closes none.

Class `Trainer` in `cell_classification/train.py` contains the training loop, loading of data and models as well as the validation callbacks. To make the training pipeline work, I needed to write some boilerplate functions in `cell_classification\augmentation_pipeline.py` make the augmentation pipeline compatible with `tf.data.dataset.map`. I also found a bug or misconception in deepcells `create_semantic_head` and opened an [issue in their repo](https://github.com/vanvalenlab/deepcell-tf/issues/624) and fixed it in `cell_classification\semantic_head.py`. It applied ReLU activation function for single channel outputs which makes no sense.

**Remaining issues**

None hopefully :>


cell_classification/augmentation_pipeline.py
```diff
@@ -90,3 +91,49 @@ def get_augmentation_pipeline(params):
         ]
     )
     return augmentation_pipeline
+
+
+def prepare_tf_aug(augmentation_pipeline):
```
ngreenwald, line 96: We should add tests to at least make sure these functions run. If it's easy to do we can check correctness, but making sure they take an `augmentation` object and don't error would be a good minimum

cell_classification/configs/params.toml
```diff
@@ -0,0 +1,32 @@
...
+gaussian_blur_prob = 0.5
+gaussian_blur_min = 0.05
+gaussian_blur_max = 0.15
+contrast_prob = 0.5
+contrast_min = 0.8
+contrast_max = 1.2
+raw_keys = ["mplex_img"]
+label_keys = ["binary_mask", "marker_activity_mask"]
+batch_size = 4
+[classes]
```
ngreenwald, line 29: What's this for?

cell_classification/configs/params.toml
```diff
@@ -0,0 +1,32 @@
...
+gaussian_noise_min = 0.05
+gaussian_noise_max = 0.15
+gaussian_blur_prob = 0.5
+gaussian_blur_min = 0.05
+gaussian_blur_max = 0.15
+contrast_prob = 0.5
+contrast_min = 0.8
+contrast_max = 1.2
+raw_keys = ["mplex_img"]
+label_keys = ["binary_mask", "marker_activity_mask"]
```
ngreenwald: The binary mask is an input, not a label, right?
Or is this describing the data type, not the relationship between inputs/labels

cell_classification/train.py
```diff
@@ -0,0 +1,149 @@
...
+    """Builds, trains and writes validation metrics for models"""
+
+    def __init__(self, params):
+        """Initialize the trainer with the parameters from the config file
+        Args:
+            params (dict): Dictionary of parameters from the config file
+        """
+        self.params = params
+
+    def prep_data(self):
```
ngreenwald, line 25: Is there a `seed` we can set for the data_prep step? We want the dataset splits to be identical for the same input data if we re-run it when testing other model params. The other option would be to do this ahead of time and save the splits.

ngreenwald: Nevermind, this is already deterministic since it's just taking from the top of record

cell_classification/train.py
```diff
@@ -0,0 +1,149 @@
...
+        self.train_dataset = self.train_dataset.map(self.prep_batches)
+        self.validation_dataset = self.validation_dataset.batch(self.params["batch_size"])
+        self.validation_dataset = self.validation_dataset.map(self.prep_batches)
+
+    def prep_model(self):
+        """Prepares the model for training"""
+        self.optimizer = Adam(learning_rate=self.params["lr"], clipnorm=0.001)
+        self.lr_sched = rate_scheduler(lr=self.params["lr"], decay=0.99)
+
+        if "test" in self.params.keys() and self.params["test"]:
```
ngreenwald, line 65: Since we may try out different models, probably makes more sense to have unique names for each, that way we can just add on to them.

cell_classification/train.py
```diff
@@ -0,0 +1,149 @@
...
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--params", type=str, default="cell_classification/configs/params.toml",
+    )
+    args = parser.parse_args()
+    params = toml.load(args.params)
+    trainer = Trainer(params)
+    trainer.train()
```
ngreenwald, line 162: Did it train successfully? How far did you get in testing things out?

ngreenwald (changes requested):
Great start! Just some minor cleanup comments.

cell_classification/configs/params.toml
JLrumberger: Deepcells `PanopticNet` expects a dictionary for the argument `num_semantic_classes` that maps output keys to output channels, i.e. {"marker_positive":1} which is what params["classes"] when it's written like this in the toml.

cell_classification/configs/params.toml
JLrumberger: This is for the data augmentation pipeline only and there it gives only the datatype. I guess I should rename it to `float_keys` and `int_keys`.

cell_classification/train.py
JLrumberger: Yep, but we should probably think about more nuanced ways for data splitting a bit later in the project.

cell_classification/train.py
JLrumberger: Good idea. I just added this single layer test model here so that we can test if the training loop works. I think we'll probably have a function that takes the params dict and returns a model at a later stage.

cell_classification/train.py
JLrumberger: Had so much other stuff to do.. I'll deploy it tomorrow to our cluster :|

ngreenwald: No worries, I wasn't sure if you had already tested it, or if it's possible there's still some small bugs

cell_classification/configs/params.toml
JLrumberger: I just found that we don't use these keys anyways, since I hard coded the logic in `augmentation_pipline.py/py_aug` and `prep_tf_aug`.

cell_classification/train.py
JLrumberger: I'd add such a function at a later stage.

JLrumberger:
Build still fails even though I tried a lot of stupid stuff. I'll look into this tomorrow again.. It's almost funny that I don't have a problem running this build in a conda environment that pretty much ignores the exact versions of the dependencies..

ngreenwald:
That's very frustrating. It's probably conflicts between `ark` and `deepcell`, since they use a lot of the same underlying packages. I don't think we import very much stuff from `ark`, might be worth just copying over those functions for now to save yourself the headache.

JLrumberger:
OMG! The build passed. How I solved it: I had a working conda environment and looked up the versions that almighty conda used to build the env. Before, I read [this article ](https://medium.com/knerd/the-nine-circles-of-python-dependency-hell-481d53e3e025) looked at the dependency tree and tried to figure it out myself, which was doomed to fail..

JLrumberger:
I will add more tests for the CLI in `plot_utils.py` in a future PR (see issue #17)  and will exclude `script.py` and `utils.py`  from coverage testing.

ngreenwald:
No problem. We can merge it in even if coveralls is "failing" as long as we look at the specific lines of code and agree they don't need to be tested right now

JLrumberger:
I did a bit of profiling of the data input pipeline. Here are the results for future ref:

batch_size 8
0. 121ms/step
1. 111ms/step with num_parallel_calls=tf.data.AUTOTUNE in every map function
2. 106ms/step 1. + prefetch(tf.data.AUTOTUNE) last -> 13.5ms/im

batch_size 4
2. 74ms/step -> 18.5ms/im

batch_size 16
2. 188ms/step -> 11.875ms/im

batch_size 24
2. 285ms/step -> 11.875ms/im

batch_size 32
2. 375ms/step -> 11.718ms/im

batch_size 64
2. 741ms/step -> 11.578ms/im

ngreenwald (approved):
Nice work!

Issue #19, opened 2022-09-16
Evaluate models / calculate scores for the pixel level


JLrumberger:
### Instructions
Write code to do predictions given a model checkpoint and calculate performance metrics for the predictions.

**Relevant background**

We trained a model on the TNBC tfrecord dataset and need to calculate pixel-wise and cell-wise evaluation metrics. This issue concerns pixel-wise metrics, but they should be implemented such that metric calculation also works for cell-wise data to re-use code.

**Design overview**

1. Write prediction routine that loads model weights, takes images and returns predictions. This should also include tile & stitch functionality.
2. Calculate pixel-level ROC curves. Find best threshold based on ROC and report confusion matrix, accuracy, precision and recall for all pixels. Report these metrics for each marker individually. Here we assume that we have a dict for each predicted sample with the keys ["instance_mask", "imaging_platform", "marker_activity_mask", "dataset", "marker", "activity_df", "prediction"]. We iterate through the sample dicts and calculate ROC curves for each single one and store the ROC values in a list. Then we can average the ROC curves and plot it together with it's standard deviation. From the mean ROC curve we get the thresholds. Iterate through the samples again and calculate confusion matrix, accuracy, precision and recall for each sample and store it in a dataframe.
3. Write a small script that loads tfrecord validation data, uses prediction routine from 1. and calculates metrics from 2. 

**Code mockup**
1.
```
class Pred
  def init(params)
    self.model = load_model()
  def predict(self, x):
    if img > params["input_shape]:
      y_hat = self.tile_and_stitch(x)
    else:
      y_hat = self.model(x)
    return y_hat
  def tile_and_stitch(x)
    ...
    return y_hat
```
2.
```
def plot_avg_roc_from_list(list)
    return plot, mean_roc

def calc_metrics(y_hat, y_true, threshold)
    y_hat_binary = (y_hat > threshold).astype(int)
    sample_df = {
    tp,fp,tn,fn = confusion_matrix(y_hat, y_true)
    precision = tp/(tp+fp)
    recall = tp/(tp+fn)
    f_score = tp/(tp+0.5*(fp+fn))
    accuracy = (tp+tn)/(tp+fp+tn+fn)
    return dict
```

3.
```
# load tfrecord
pred = Pred(params)
roc_list = []
for sample in record:
    y_hat = pred(sample["mplex_img"])
    save(y_hat)
    roc_list.append(sklearn.ROC(y_hat, y_true))

[...]

PR #20, opened 2022-09-19
Predict with model and calculate pixel-wise evaluation metrics

JLrumberger:

This code loads a model checkpoint and the associated `params.toml` and prepares predictions and calculates pixel-level metrics (ROC curve, AUC, precision, recall, f1_score) for the binary segmentation problem. This PR closes issue #19 

I added class functions `predict` and `load_model` to the `ModelBuilder` class and the associated tests in `model_builder_test.py`. I added script that loads a model and calculates predictions as well as evaluation metric calculations in `predict.py` and plots for these metrics in `plot_utils.py` plus associated tests in `plot_utils_test.py`.

**Remaining issues**

`tile_and_stitch` functionality is still missing in `ModelBuilder` and I'll add it via a separate PR at a later stage. Also tests for evaluation metric calculation and plots are not really thorough, but I use sklearn functionality for most of the metric calculation and consider it tested and correct.


cell_classification/predict.py
```diff
@@ -0,0 +1,236 @@
...
+            validation data
+        params (dict):
+            dictionary containing model and validation data paths
+    Returns:
+        predictions (list):
+            list of dictionaries containing the full example + prediction
+        params (dict):
+            dictionary containing model hyperparams and validation data paths
+    """
+    pred_list = []
```
ngreenwald: Is this eventually going to be a problem keeping all of the validation set in memory?

cell_classification/predict.py
```diff
@@ -0,0 +1,236 @@
...
+    for sample in tqdm(val_dset):
+        sample["prediction"] = model.predict(model.prep_batches(sample)[0])
+        pred_list.append(sample)
+
+    # split batches to single samples
+    single_example_list = []
```
ngreenwald: Is this because the tf_record is constructed with a specific batch_size already?

cell_classification/predict.py
```diff
@@ -0,0 +1,236 @@
...
+            key for predictions
+    Returns:
+        roc (dict):
+            dictionary containing ROC curve data
+    """
+    roc = {"fpr": [], "tpr": [], "thresholds": [], "auc": []}
+    for sample in pred_list:
+        if sample[gt_key].max() == 0:
+            continue
+        fpr, tpr, thresholds = roc_curve(sample[gt_key].flatten(), sample[pred_key].flatten())
```
ngreenwald: This will need to be tweaked for cell-level metrics. Do you think the best thing is just get the pixel-level version working first, and then worry about cell-level in a different PR? Or try and do both here?

cell_classification/predict.py
```diff
@@ -0,0 +1,236 @@
...
+        metrics["threshold"] = threshold
+        for key in ["dataset", "imaging_platform", "marker"]:
+            metrics[key] = sample[key]
+        return metrics
+
+    thresholds = np.linspace(0, 1, 101)
+    metric_list = Parallel(n_jobs=8)(delayed(_calc_metrics)(i) for i in thresholds)
+    avg_metrics = deepcopy(metrics_dict)
+    for key in ["dataset", "imaging_platform", "marker", "threshold"]:
+        avg_metrics[key] = []
```
ngreenwald: In general this file has pretty dense code without many comments

ngreenwald (changes requested):
Some minor comments below. Overall looks quite good, but it's a bit harder for me to give high-level feedback once the code is already written, more helpful to do it at the design doc stage. Especially as we think about the best way to integrate pixel-level, cell-level, and multi-marker cell-level metrics, I think doing some planning ahead of time will make things easier.

cell_classification/predict.py
JLrumberger: When the dataset goes 100x it will be a problem. I'll serialize results in this function individually, probably via .hdf instead of .pkl then it can be loaded with fiji.

cell_classification/predict.py
JLrumberger: No, the dataloader has a batchsize that we can set and change anytime, the tf_record just stores single examples. I did this because we want to do inference on larger batches. Atm with batchsize=4 my laptop GPU is far from fully utilized and it runs wit 5-7 batches per second.

cell_classification/predict.py
JLrumberger: I think small stuff like the above should be changed in this PR, but I'd like to do cell level stuff in a another PR since this one contains already a lot of changes.

cell_classification/predict.py
ngreenwald: so when you run `model.predict`, that is predicting 4 images at a time. Then the logic from line 56 onwards is splitting that batch back into individual examples?

cell_classification/predict.py
ngreenwald: Got it, sounds good

cell_classification/predict.py
ngreenwald: Sounds good!

cell_classification/predict.py
JLrumberger: Yes exactly. My rationale was that we'll apply this model on larger images anyways and need to do tile & stitch for prediction, so it'll almost always run on a batch of tiles and not on single tiles.

cell_classification/predict.py
JLrumberger: I'll leave the `.flatten()` here as it is, since it does nothing if the array is already flat which will be the case for cell-level data.

cell_classification/predict.py
JLrumberger: I added more comments and to make it more understandable.

JLrumberger:
@ngreenwald I requested another review but I think I'll just implement the cell level stuff in this branch and then push it again. Cell level stuff seems to be less work than expected. I let you know once I am done with it so that you can review it.

ngreenwald (changes requested):
Sounds good, also subsetting the pixels to be the relevant ones

JLrumberger:
This one took a while, because I had no errors when running it locally and it was a bit harder to find the error.

cell_classification/predict.py
```diff
@@ -0,0 +1,375 @@
...
+        mask = sample["instance_mask"] == unique_label
+        mean_pred = sample["prediction"][mask].mean()
+        df.at[df.index[df["labels"] == unique_label].to_list().pop(), "pred_activity"] = mean_pred
+        mean_per_cell_mask[mask] = mean_pred
+    sample["activity_df"] = df
+    sample["prediction_mean"] = mean_per_cell_mask
+    return sample
+
+
+def predict(model, val_dset, params, return_pred=False):
```
ngreenwald: Tons of new code, which is great. Let's talk about it tomorrow after standup, I think it'll be easier to go through this in person. 

The main thing I want to discuss: what data structures are we using for each step of the model evaluation pipeline?

JLrumberger:
Hey @ngreenwald,
yesterday we discussed if I could wrap the model with the [deepcell application class](https://github.com/vanvalenlab/deepcell-tf/blob/master/deepcell/applications/mesmer.py) and also use [their metrics](https://github.com/vanvalenlab/deepcell-toolbox/blob/master/deepcell_toolbox/metrics.py) and include both into this PR.

I thought a bit about that. For the training and model evaluation part we can assume that all datasets the model interacts with are constructed using our tfrecord dataset construction class. I'd further assume that we can calculate metrics on tiles and don't need to stitch images before calculating metrics. Because of this, I think this PR should just be about the validation and testing of the training pipeline and the model. Thus it needs to include a prediction routine and metric calculation. But it doesn't need to include the application layer for endusers and I'd rather have this in a separate PR. Do you agree?

cell_classification/predict.py
```diff
@@ -0,0 +1,375 @@
...
+    """
+    params["eval"] = True
+    model = ModelBuilder(params)
+    model.prep_data()
+    model.load_model(params["model_path"])
+    val_dset = iter(model.validation_dataset)
+    return model, val_dset
+
+
+def process_to_cells(sample):
```
ngreenwald: This can be split into two functions
1) A general purpose function which computes the average value of a probability mask within each of the segmented objects provided in a segmentation mask. It returns the results in a pandas df
2) A metrics-specific function which appends the predicted activity next to the ground-truth activity in the activity df

cell_classification/predict.py
```diff
@@ -0,0 +1,375 @@
...
+
+    def __next__(self):
+        if self.file_idx >= len(self.files):
+            raise StopIteration
+        else:
+            self.file_idx += 1
+            return self.load_hdf(self.files[self.file_idx - 1])
+
+
+if __name__ == "__main__":
```
ngreenwald: Lets try and keep the code for running evaluations separate from the library of defined functions. This can go in `generate_accuracy_predictions.py` or something like that

ngreenwald (changes requested):
Looks great, just a couple tweaks

cell_classification/predict.py
JLrumberger: Yep, sounds good.

cell_classification/post_processing.py
```diff
@@ -0,0 +1,51 @@
...
+        pd.DataFrame:
+            DataFrame of cell level predictions
+    """
+
+    unique_labels = np.unique(instance_mask)
+    unique_labels = unique_labels[unique_labels != 0]
+    mean_per_cell_mask = np.zeros_like(instance_mask, dtype=np.float32)
+    df = pd.DataFrame(columns=['labels', 'pred_activity'])
+    i = 0
+    for unique_label in unique_labels:
```
ngreenwald, line 24: This can be runtime optimized once we're working with a bigger dataset, for now it's fine.

ngreenwald (approved):

Issue #21, opened 2022-10-13
Training tweaks (grokking training scheme, selective loss function and label smoothing) 


JLrumberger:
### Instructions
In order to get a good baseline model a few tweaks need to be implemented:
- improve optimization by adding [grokking training scheme](https://arxiv.org/abs/2201.02177) (cosine decaying learning rate scheduler, weight decay) to model_builder.py. 
- Add `loss_function.py` where we have a loss_function class that wraps `tf.keras.losses` and has a class method that let's us set the loss to zero for image regions whose training label is -1 (i.e. not specific enough marker for the celltype)

**Relevant background**

This just adds a few tweaks that are known to improve training and generalization. It should give us a good baseline to compare future approaches that target the noisy labels problem. 

**Design overview**

- Weight decay is a class attribute of tf.keras.layers, so I'll write a utility function that takes in a model, iterates through the layers and adds l2 weight decay on kernels and biases of all layers that contain weights. 
- LR scheduler : we use the one from `deepcell.utils.train_utils.rate_scheduler` at the moment. If deepcell has a cosine decay scheduler we'll use this one, if not I'll replace it with `tf.keras.optimizers.schedules.CosineDecay` in `prep_model.py` (l64)
- I'll replace `ModelBuilder.prep_loss` with a separate class that wraps `tf.keras.losses` and zeros out loss regions where the label is -1, it should also allow to use other loss functions from `tf.keras.losses` such as focal loss

**Code mockup**

```
LossFunction():
  def __init__(self, loss_name: str, selective_masking: bool, kwargs*)
    self.loss_fn = self.get_loss_fn(loss_name, kwargs*)

  def mask_loss(self, loss_img, gt):
    loss_img[gt==-1] = 0

  def get_loss_fn(self, loss_name):
    self.loss_fn = getattr(tf.keras.losses, loss_name)(kwargs*)

  def forward(pred, gt):
    loss_img = self.loss_fn(pred, gt)
    if self.selective_masking:
      loss_img = self.mask_loss(loss_img, gt)
    return loss_img.mean()
```
**Required inputs**

- Weight decay requires a factor to control its strengths (i.e. 1e-4)
- LR scheduler requires minimal LR and training steps
- loss functions use different arguments and take kwargs* so that they are all accessible via this wrapper

**Output files**

This just improves training but no direct output is created.

**Timeline**
Give a rough estimate for how long you think the project will take. In general, it's better to be too conservative rather than too optimistic. 
- [X] A couple days
[...]

PR #22, opened 2022-10-14
Training tweaks and loss wrapper

JLrumberger:

This PR closes Issue #21 by adding grokking training scheme, selective masking of the loss and a wrapper for loss functions from `tf.keras.losses` that allows to use label-smoothing.  

Grokking training scheme:
- added weight decay to model_builder.py
- added cosine annealing learning rate scheduler

Loss masking and label smoothing:
- added a loss wrapper in loss.py that implements selective masking and uses tf.keras.losses loss functions instead of the deepcell ones. The keras losses allow label smoothing by default. 

**Remaining issues**

None

### DIFF
--- a/cell_classification/configs/params.toml
+++ b/cell_classification/configs/params.toml
@@ -26,5 +26,12 @@ contrast_prob = 0.5
 contrast_min = 0.8
 contrast_max = 1.2
 batch_size = 4
+steps_per_epoch = 100
+loss_fn = "BinaryCrossentropy"
+loss_selective_masking = true
+[loss_kwargs]
+from_logits = true
+label_smoothing = 0.1
+
 [classes]
 marker_positive = 1
--- a/cell_classification/loss.py
+++ b/cell_classification/loss.py
@@ -1,0 +1,63 @@
+import tensorflow as tf
+from tensorflow.keras import backend as K
+
+
+class Loss():
+    """ Wrapper for loss functions that allows for selective masking of the loss
+    """
+    def __init__(self, loss_name, selective_masking, **kwargs):
+        """ Initialize the loss function
+        Args:
+            loss_name (str):
+                name of the loss function
+            selective_masking (bool):
+                whether to use selective masking
+            **kwargs:
+                additional arguments for the loss function
+        """
+        self.loss_fn = getattr(tf.keras.losses, loss_name)(
+            reduction=tf.keras.losses.Reduction.NONE, **kwargs
+        )
+        self.selective_masking = selective_masking
+
+    def mask_out(self, loss_img, y_true):
+        """ Selectively mask the loss by setting it to zero where y_true == -1
+        Args:
+            loss_img (tf.Tensor):
+                loss image
+            y_true (tf.Tensor):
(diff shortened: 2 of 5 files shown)


cell_classification/loss.py
```diff
@@ -0,0 +1,63 @@
...
+            loss_img (tf.Tensor):
+                loss image
+            y_true (tf.Tensor):
+                ground truth image
+        Returns:
+            tf.Tensor:
+                masked loss image
+        """
+        y_true = tf.reshape(y_true, tf.shape(loss_img))
+        return tf.where(y_true == 2, tf.zeros_like(loss_img), loss_img)
```
ngreenwald, line 35: how come we're checking if `y_true==2`?

cell_classification/configs/params.toml
```diff
@@ -26,5 +26,12 @@ contrast_prob = 0.5
 contrast_min = 0.8
 contrast_max = 1.2
 batch_size = 4
+steps_per_epoch = 100
+loss_fn = "BinaryCrossentropy"
+loss_selective_masking = true
+[loss_kwargs]
+from_logits = true
+label_smoothing = 0.1
```
ngreenwald, line 34: It doesn't look like label_smoothing is ever popped from the params. Is this handled automatically somewhere?

ngreenwald: I assume that with only two classes, this won't have a huge impact, but given how easy it is to implement probably worth including?

ngreenwald (changes requested):

cell_classification/loss.py
JLrumberger: I replaced the -1 for unspecific markers with 2 in order to store the groundtruth mask in uint8 instead of int32.

cell_classification/model_builder.py
```diff
@@ -128,19 +138,17 @@ def train(self):
...
-        if n_classes == 1:
-            def loss_fn(y_true, y_pred):
-                return tf.keras.losses.binary_crossentropy(y_true, y_pred, from_logits=False)
-        elif n_classes > 1:
-            def loss_fn(y_true, y_pred):
-                return tf.keras.losses.categorical_crossentropy(y_true, y_pred, from_logits=False)
+        loss_fn = Loss(
+            self.params["loss_fn"], self.params["loss_selective_masking"],
+            **self.params["loss_kwargs"]
+        )
```
JLrumberger, line 151: Here we pop params['loss_kwargs]

cell_classification/configs/params.toml
JLrumberger: Not sure if I understand this correctly. So under `[loss_kwargs]` you can put all arguments that you want to put into the init of your tf.keras.loss. In [model_builder.py](https://github.com/angelolab/cell_classification/blob/8b7816cbb8fe041c34c40402d9949e73f6705159/cell_classification/model_builder.py#L150)  we pop `loss_kwargs` from params and it is used by the keras loss function to modify loss calculation.

ngreenwald (approved):

PR #27, opened 2022-10-29
Added manual training loop to capture training data within epochs

JLrumberger:

I replaced the `keras.fit` training loop with a custom training loop in `model_builder.py` in order to write out snapshots and training/validation loss every n steps instead of every n epochs. 

Added `train_step()` and `distributed_train_step()` to `ModelBuilder` and changed `train()`

**Remaining issues**

I need to do further tests to see if everything works fine for multi-GPU training. At the moment the validation loss is weirdly scaled for multi - gpu training.


JLrumberger:
I'll open a new issue for checking if everything works with multi-gpu training since it's not a pressing problem.

ngreenwald (approved):
Looks good, agreed re: saving multi-gpu for when we need it

PR #28, opened 2022-11-01
Added promix naive loss masking scheme

### LINKED ISSUE #26: Promix Naive
**Relevant background**

We assume that 5-15% of the marker positive/negative labels in our dataset are wrong. To cope with this, we will adapt the clean sample selection procedure from [ProMix](https://arxiv.org/pdf/2207.10276.pdf) from the image-level to the cell-level. 

**Design overview**

We adapt the methods presented as ProMix Naive for the binary pixel-wise classification case as follows:
1. Class-wise small-loss selection:
- after each epoch they split the training data into subsets based on its training label, s.t. one has C subsets S for each label c_0,c_1,..., c_C. They calculate the CE loss for the samples in each subset and select for each subset the top-k samples with the lowest loss as their clean training data, thus arriving at a roughly balanced sample
- We'll calculate the running 5%-percentile of the loss on foreground GT areas, separately for GT positive and negative areas. Then we'll mask out the loss where it is higher than the running 5% percentile of each class

2. Matched High-Confidence Selection
- they calculate confidence scores for each training sample and sample the ones where the confidence is higher than a given threshold tau (tau = 0.99 for CIFAR10N and tau = 0.95 for CIFAR100N)
- We mask out areas in predictions where the predicted confidence is lower than the threshold tau = 0.95

The loss and confidence scores used in the above calculations are based on **un-augmented** training data

3. Consistency Regularization
- they augment the images with weak augmentations (zoom/crop, color jitter, gaussian blur) and strong augmentations (AutoContrast, Equalize, Rotate, Solarize, Color, Posterize, Contrast, Brightness, Sharpness, Shear, Translate) and calculate the loss based on the predictions and the GT. So this is **not augmentation consistency training** but just normal data augmentation. 
- I'll just keep my augmentation pipeline as is, since I don't think its much different

4. Mixup
- they use mixup with sigma~Beta(4,4) 
- we'll just use the [segmentation version of mixup](https://openreview.net/pdf?id=rkBBChjiG)

**Code mockup**

For 1. and 2. we'll sub-class `ModelBuilder` and change class functions 
- `prep_data`: exclude mapping the data augmentation onto the train dataset
- `prep_model`: exclude `model.compile` from here
- `train`:  write custom training loop that looks like the one below

add class functions
- `class_wise_loss_selection(y_pred, y_gt)`:  calculates k-th percentile of GT positive and negative CE loss per class and stores them via exponential-moving-average in a class attribute, returns a loss mask based on this 
- `matched_high_confidence_selection(y_pred, y_gt)`: returns a loss mask for all predictions whose confidence are below a certain threshold tau

```
for x,y_gt in data:
    y_pred = model(x)
    loss_mask = model.class_wise_loss_selection(y_pred, y_gt)
    loss_mask += model.matched_high_confidence_selection(y_pred, y_gt)
    x_aug, loss_mask_aug, y_gt_aug = aug_fn(x, loss_mask, y_gt)
    with tf.gradient_tape as tape:
        y_pred_aug = model(x_aug)
        loss_value = model.loss_fn(y_gt_aug, y_pred_aug)
        loss_value = model.masked_loss_fn(loss_value, loss_mask_aug)

    grads = tape.gradient(loss_value, model.trainable_weights)
    optimizer.apply_gradients(zip(grads, model.trainable_weights))
```
Code inside the training loop should be put into a static class function decorated by `@tf.function` in order to use static-graph mode during training. Probably we need to change the augmentation library we're using, this could be a good time to switch to one of the tf-based augmentation libraries.

**Required inputs**

Provided a description of the required inputs for this project, including column names for dfs, dimensions for image data, prompts for user input, directory structure for loading data, etc

**Output files**

Provide a description of the outputs for this project. If any plots will be generated, provide (simple) sketches demonstrating the plot type and axes labels. 

**Timeline**
- [X] A couple days
- [ ] A week

JLrumberger:

This PR adds promix naive as described in #26, but without data augmentation. I'd add a new data augmentation pipeline in a separate PR, since this one has already lots of changes.

As described in the design doc #26 I sub-classes ModelBuilder, replaced class functions `prep_data`, `prep_model `, `train`, `train_step` and added the following class functions:
- `matched_high_confidence_selection_thresholds`, that converts confidence thresholds to loss thresholds
- `reduce_to_cells` that calculates mean-per-cell predictions from semantic segmentation maps
- `batchwise_loss_selection` which iterates over the batch and calculates the `loss_mask` for each sample
- `class_wise_loss_selection` which implements the class-wise loss selection procedure from promix
- `matched_high_confidence_selection` which implements the matched high-confidence selection from promix. Here we do it based on the loss instead of the confidence for shorter runtime.

In class `ModelBuilder` I disentangled `tensorboard_callbacks` from the training loop into its own class function to re-use it in `PromixNaive`.

**Remaining issues**

Need to add data augmentation and MixUp as stated in issue #26

### DIFF
--- a/cell_classification/configs/params.toml
+++ b/cell_classification/configs/params.toml
@@ -30,6 +30,7 @@ loss_fn = "BinaryCrossentropy"
 loss_selective_masking = true
 quantile = 0.3
 confidence_thresholds = [0.1, 0.9]
+ema = 0.01
 [loss_kwargs]
 from_logits = false
 label_smoothing = 0.1
--- a/cell_classification/configs/params.toml
+++ b/cell_classification/configs/params.toml
@@ -28,6 +28,8 @@ contrast_max = 1.2
 batch_size = 4
 loss_fn = "BinaryCrossentropy"
 loss_selective_masking = true
+quantile = 0.3
+confidence_thresholds = [0.1, 0.9]
 [loss_kwargs]
 from_logits = false
 label_smoothing = 0.1
--- a/cell_classification/metrics_test.py
+++ b/cell_classification/metrics_test.py
@@ -7,6 +7,7 @@ import os
 from metrics import calc_roc, calc_metrics, average_roc, HDF5Loader
 import numpy as np
 import pandas as pd
+import tensorflow as tf
 
 
 def make_pred_list():
@@ -97,17 +98,13 @@ def test_HDF5Generator():
         params["record_path"] = tf_record_path
         params["path"] = temp_dir
         params["experiment"] = "test"
-        params["num_epochs"] = 0
+        params["num_steps"] = 2
         params["num_validation"] = 2
-        params["snap_steps"] = 10
-        params["val_steps"] = 10
+        params["snap_steps"] = 100
+        params["val_steps"] = 100
         model = ModelBuilder(params)
         model.train()
-        model.params["eval"] = True
-        model.prep_data()
-        val_dset = iter(model.validation_dataset)
-
-        single_example_list = model.predict_dataset(val_dset, save_predictions=True)
+        model.predict_dataset(model.validation_dataset, save_predictions=True)
         generator = HDF5Loader(model.params['eval_dir'])
 
         # check if generator has the right number of items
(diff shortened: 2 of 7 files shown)


ngreenwald (approved):
Great work!

PR #31, opened 2022-11-07
Added exclude_background_tiles and fixed normalization

JLrumberger:

This PR changes the `SegmentationTFRecords` class:
1. quantile normalization now only calculates quantiles on pixels whose intensity is > 0, so that sparse markers don't get super dim because of normalization
2. we add an argument `exlude_background_tiles`  to the class, if it's set to true, only tiles that contain foreground in the label mask (i.e. 1s or 2s) are stored in the tfrecord. All tiles that only contain background (0s) are not saved in the tfrecord.

Changed a few lines in `segmentation_data_prep.py`

**Remaining issues**

None

### DIFF
--- a/cell_classification/segmentation_data_prep.py
+++ b/cell_classification/segmentation_data_prep.py
@@ -20,7 +20,7 @@ class SegmentationTFRecords:
         tile_size, stride, tf_record_path, selected_markers=None, normalization_dict_path=None,
         normalization_quantile=0.999, cell_type_key="cluster_labels", sample_key="SampleID",
         segmentation_fname="cell_segmentation", segment_label_key="labels",
-        segmentation_naming_convention=None
+        segmentation_naming_convention=None, exlude_background_tiles=False
     ):
         """Initializes SegmentationTFRecords and loads everything except the images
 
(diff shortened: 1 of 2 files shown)


cell_classification/segmentation_data_prep.py
```diff
@@ -465,7 +472,9 @@ def calculate_normalization_matrix(self, data_folders, selected_markers):
                 img = self.get_image(data_folder, marker)
                 if marker not in quantiles:
                     quantiles[marker] = []
-                quantiles[marker].append(np.quantile(img, self.normalization_quantile))
+                foreground = img[img > 0]
+                if np.sum(foreground) > 0:
+                    quantiles[marker].append(np.quantile(foreground, self.normalization_quantile))
```
ngreenwald, line 477: Did you check to see if this resolves the issue with the poor contrast? If not, we should open an issue about it and can later track it down.

ngreenwald (changes requested):

cell_classification/segmentation_data_prep.py
JLrumberger: CD20:
Old:
download
New
download

SMA:
Old:
download

New:
download

JLrumberger: Looks better I'd say. Most of the salt&pepper-like noise is reduced a lot.

ngreenwald: Agreed!

ngreenwald (approved):
Looks good

PR #33, opened 2022-11-07
Data aug

### LINKED ISSUE #26: Promix Naive

JLrumberger:

This PR closes #26 by adding the tf augmentation pipeline and mixup.

Added classes `Flip`, `Rot90`, `GaussianNoise`, `GaussianBlur`, `Zoom`, `LinearContrast`, `MixUp` and `Augmenter`. Augmenter applies augmentations on each image in a batch individually. MixUp works directly on batches, thus shouldn't be passed to Augmenter.

**Remaining issues**

None


cell_classification/augmentation_pipeline.py
```diff
@@ -137,3 +138,351 @@ def py_aug(batch, tf_aug):
...
+        if tf.random.uniform(()) < self.prob:
+            zoom_factor = tf.random.uniform([1], self.min_zoom, self.max_zoom)
+            zoom_factor = 1/zoom_factor
+            zooms = tf.expand_dims(
+                tf.cast(
+                    tf.concat([zoom_factor, zoom_factor], axis=0), dtype=tf.float32
+                ), 0
+            )
+            squeeze = False
+            if tf.rank(image) == 3:
```
ngreenwald, line 300: Are all of the channels currently stored as row x col x 1 3D images, 1 x row x col x 1 4D images, or does it depend on how they were saved?

cell_classification/augmentation_pipeline.py
```diff
@@ -137,3 +138,351 @@ def py_aug(batch, tf_aug):
...
+            tf.Tensor:
+                The augmented image batch
+            tf.Tensor:
+                The augmented label batch
+        """
+        labels = tf.cast(labels, dtype=tf.float32)
+        undecided = tf.where(labels == 2, True, False)
+        b = tf.shape(image)[0]
+        prob = tf.random.uniform([b], 0, 1) < self.prob
+        beta = tf.random.uniform([b], 0, 1)
```
ngreenwald, line 397: Some more comments here would be useful to explain how the vector `b` is being used to determine which images are augmented, the transformations to get the desired distribution, etc

cell_classification/augmentation_pipeline.py
```diff
@@ -137,3 +138,351 @@ def py_aug(batch, tf_aug):
...
+            tf.Tensor:
+                The augmented labels
+        """
+        if tf.random.uniform(()) < self.prob:
+            factor = tf.random.uniform((), self.min_factor, self.max_factor)
+            image *= factor
+        return image, labels
+
+
+class MixUp(tf.Module):
```
ngreenwald, line 360: Using MixUp, even without a schedule for `p` for promix, will help to avoid the issue we saw of the predictions being heavily skewed towards 0 and 1. Even with only high-confidence positives and negatives in the original training data, we'll now have targets that are in the middle regime

cell_classification/promix_naive.py
```diff
@@ -69,11 +72,15 @@ def train_step(model, optimizer, loss_fn, aug_fn, loss_mask, x, y_gt):
         Returns:
             tf.Tensor: loss value
         """
-        x_aug, loss_mask_aug, y_gt_aug = aug_fn(x, loss_mask, y_gt)
+        loss_mask = tf.expand_dims(tf.cast(loss_mask, tf.uint8), -1)
+
+        x_aug, cat = aug_fn(x, tf.concat([loss_mask, y_gt], axis=-1))
+        loss_mask_aug, y_gt_aug = tf.split(cat, [loss_mask.shape[-1], y_gt.shape[-1]], axis=-1)
```
ngreenwald: Couldn't follow this logic

ngreenwald (changes requested):
Looks great, just some small comments. Did you already check the outputs for the whole pipeline? Can you share the images once you do? I'm curious to see what the range of augmentations is for 1) the non-mixup part of the pipeline and 2) the mixup part of the pipeline. It would be great if you could take a single image, and run it through the pipeline a bunch of different times to get a sense for what the range of modifications is. If you'd rather do that later, not part of this PR, that's totally fine

cell_classification/augmentation_pipeline.py
JLrumberger: The `tf.map` function in augmenter squeezes the batch dimension. Since we'll probably use the augmentations only via `Augmenter` and thus `tf.map` I could get rid of the control flow and just assume HWC everywhere inside the augmentations.

cell_classification/augmentation_pipeline.py
JLrumberger: 100%, the mixup function is not really readable atm.

JLrumberger:
Non mixup part
image
Without augmentations, with augmentations and with MixUp
image

In the promix paper they parameterize like this Beta(4.0,4.0), thus their actual \sigma \in [0.84, 1] which is used to weight image1 and image2 in MixUp. Honestly I don't see how MixUp improves results yet, qualitatively the images just look weird with it.

cell_classification/promix_naive.py
JLrumberger: `aug_fn` applies non-spatial augmentations (noise, blur, contrast) and bilinear interpolation for the spatial augmentations only for the first argument, for the second argument nearest neighbor interpolation is used and no non-spatial augmentations. Since we want noise/blur/contrast only on the multiplexed image, we split and concat stuff around here..

cell_classification/prepare_data_script.py
```diff
@@ -25,7 +25,7 @@ def naming_convention(fname):
     normalization_dict_path=os.path.normpath(
        "C:/Users/lorenz/Desktop/angelo_lab/TONIC/normalization_dict.json"
     ),
-    normalization_quantile=0.99,
+    normalization_quantile=0.999,
```
ngreenwald, line 28: Will you plot a couple side-by-side, along with the raw image scaled 0-1 to max, to see what the effects are?

cell_classification/augmentation_pipeline.py
```diff
@@ -130,10 +133,373 @@ def py_aug(batch, tf_aug):
...
+        beta = tf.math.maximum(beta, 1 - beta)
+        beta = tf.math.pow(beta, 1 / self.alpha)
+        beta = tf.where(prob, beta, tf.ones_like(beta))
+        prob = tf.cast(tf.reshape(prob, [b, 1, 1, 1]), tf.float32)
+        # apply mixup on data
+        beta = tf.reshape(beta, [b, 1, 1, 1])
+        x_mplex = beta * x_mplex + (1 - beta) * tf.reverse(x_mplex, axis=[0])
+        x_binary = beta * x_binary + (1 - beta) * tf.reverse(x_binary, axis=[0])
+        labels = beta * labels + (1 - beta) * tf.reverse(labels, axis=[0])
+        # loss mask should be 1 where both mixup loss masks == 1 and zero otherwise
```
ngreenwald, line 407: Is this the new way you're handling 2s? How come both == 1 is the correct approach for loss masks?

cell_classification/promix_naive.py
```diff
@@ -22,7 +23,42 @@ def __init__(self, params):
...
         self.loss_fn = self.prep_loss()
         self.quantile = self.params["quantile"]
+        self.quantile_start = self.params["quantile"]
+        self.quantile_end = self.params["quantile_end"]
+        self.quantile_warmup_steps = self.params["quantile_warmup_steps"]
         self.class_wise_loss_quantiles = {}
+        self.aug_fn = prepare_keras_aug(params, dtype=(tf.float32, tf.uint8))
+        self.mixup_fn = MixUp(prob=params["mixup_prob"], alpha=params["mixup_alpha"])
+
+    def quantile_scheduler(self, step):
```
ngreenwald, line 32: Nice, I'll be curious to see how this changes the output, I imagine it will be quite helpful

cell_classification/promix_naive_test.py
```diff
@@ -72,6 +72,7 @@ def test_train():
         params["quantile"] = 0.3
         params["ema"] = 0.01
         params["confidence_thresholds"] = [0.1, 0.9]
+        params["mixup_prob"] = 0.0
```
ngreenwald: Should we add a test with mixup enabled just to make sure it runs?

ngreenwald (changes requested):
Looks great, just a couple questions

cell_classification/augmentation_pipeline.py
JLrumberger: Would you rather do the promix weighted sum over the two loss_masks? I didn't know how to approach this and I thought the conservative solution is just masking out both areas, but then a lot of stuff could be masked out.. might be more clever to have to have it weighted then..

cell_classification/prepare_data_script.py
JLrumberger: Will do before I create the decidua dataset.

ngreenwald (approved):
Looking forward to seeing the results!

PR #36, opened 2022-11-17
Filter out sparse examples from the training data

### LINKED ISSUE #34: Filter out sparse examples
**Relevant background**

We want to further reduce our dataset. Thus we calculate the 40 / 60 / x % quantile of positive cells per tile for each marker and filter out all tiles that have less than the x-quantile of positive cells. 

**Design overview**

1. Calculate the x-quantile of positive cells for each marker
2. Filter out Examples from the training data that have less than x-quantile positive cells

**Code mockup**

Add class function `quantile_filter` to `ModelBuilder`

```
def quantile_filter(self):
    num_pos_dict = {}
    for example in self.training_dataset:
        marker = example["marker"]
        if marker not in num_pos_dict.keys():
            num_pos_dict["marker"] = []
        num_pos_dict["marker"].append([np.sum(example["activity_df"].labels==1))
    quantile_dict = {}
        for marker, pos_list in num_pos_dict.items():
            quantile_dict[marker] = np.quantile(pos_list, self.filter_quantile)
    predicate = lambda example: np.sum(example["activity_df"].labels==1) > quantile_dict[example["marker"]]
    self.train_dataset = self.train_dataset.filter(predicate)
```

**Required inputs**

`tf.data.TFRecordDataset` containing the training data.

**Output files**

`tf.data.TFRecordDataset` containing the filtered training data.

**Timeline**
- [x] A couple days
- [ ] A week

JLrumberger:

This PR closes #34 by adding function `quantile_filter` to `ModelBuilder`.

Added function `quantile_filter` to the `ModelBuilder` class that first calculates the x-quantile of positive cells per tile for each marker and then filters out the tiles whose number of positive cells is below the quantile. Also I aligned `prep_data` in `ModelBuilder` with the one that was in `PromixNaive` before. Thus we can get rid of the one in `PromixNaive` because it gets inherited from `ModelBuilder`.

**Remaining issues**

None

### DIFF
--- a/cell_classification/model_builder.py
+++ b/cell_classification/model_builder.py
@@ -18,6 +18,7 @@ import pandas as pd
 from loss import Loss
 from tqdm import tqdm
 from time import time
+import json
 
 
 class ModelBuilder:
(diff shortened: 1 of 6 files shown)


cell_classification/model_builder.py
```diff
@@ -356,6 +364,44 @@ def add_weight_decay(self):
...
+        Args:
+            dataset (tf.data.Dataset):
+                Dataset to filter
+        Returns:
+            dataset (tf.data.Dataset):
+                Filtered dataset
+        """
+        print("Filtering out sparse training examples...")
+        num_pos_dict = {}
+        for example in tqdm(dataset):
```
ngreenwald: We should append this to the tf.record somehow, or store it elsewhere, so that we only need to calculate the quantiles once, and we can use a different threshold in subsequent rounds given the same data

cell_classification/segmentation_data_prep.py
```diff
@@ -515,6 +515,7 @@ def parse_dict(deserialized_dict):
         example[key] = tf.strings.unicode_encode(
             tf.cast(deserialized_dict[key], tf.int32), "UTF-8"
         )
+        example[key] = example[key]
```
ngreenwald: hmm

ngreenwald (changes requested):
See below about optimizing quantile computation

cell_classification/model_builder.py
JLrumberger: Really good idea that saves us a couple of minutes ramp-up time everytime we'll start training.

ngreenwald (approved):
Nice, looks good

PR #37, opened 2022-11-17
Added deepcell application wrapper for model

JLrumberger:

Wraps the model with the deepcell `Application` class that does batching and tile & stitch during inference.

Added the required functions `cell_preprocess`, `cell_postprocess` and `format_output` as well as class `CellClassification` that inherits from `deepcell.Application`.

**Remaining issues**

None

### DIFF
--- a/cell_classification/application.py
+++ b/cell_classification/application.py
@@ -9,6 +9,7 @@ def cell_preprocess(image, **kwargs):
     Returns:
         np.array: processed image array
     """
+    output = np.copy(image)
     if len(image.shape) != 4:
         raise ValueError("Image data must be 4D, got image of shape {}".format(image.shape))
 
@@ -20,10 +21,10 @@ def cell_preprocess(image, **kwargs):
             norm_factor = normalization_dict[marker]
         else:
             print("Norm_factor not found for marker {}".format(marker))
-            norm_factor = np.quantile(image[..., 0], 0.999)
-        image[..., 0] /= norm_factor
-        image = image.clip(0, 1)
-    return image
+            norm_factor = np.quantile(output[..., 0], 0.999)
+        output[..., 0] /= norm_factor
+        output = output.clip(0, 1)
+    return output
 
 
 def cell_postprocess(model_output):
--- a/cell_classification/application.py
+++ b/cell_classification/application.py
@@ -1,0 +1,66 @@
+import numpy as np
+from deepcell.applications import Application
+
+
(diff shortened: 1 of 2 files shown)


cell_classification/application.py
```diff
@@ -0,0 +1,67 @@
...
+    normalize = kwargs.get('normalize', True)
+    marker = kwargs.get('marker')
+    normalization_dict = kwargs.get('normalization_dict')
+    if normalize:
+        if marker in normalization_dict.keys():
+            norm_factor = normalization_dict[marker]
+        else:
+            print("Norm_factor not found for marker {}".format(marker))
+            norm_factor = np.quantile(output[..., 0], 0.999)
+        output[..., 0] /= norm_factor
```
ngreenwald, line 27: Any reason we need to index into the channels dimension?

cell_classification/application_test.py
```diff
@@ -0,0 +1,35 @@
+from application import CellClassification
+from model_builder import ModelBuilder
+from segmentation_data_prep_test import prep_object_and_inputs
+import numpy as np
+import tempfile
+import toml
+import os
+
+
+def predict_test():
```
ngreenwald, line 10: We should have a separate test for the normalize function

cell_classification/application.py
```diff
@@ -0,0 +1,67 @@
...
+        raise ValueError("Image data must be 4D, got image of shape {}".format(image.shape))
+
+    normalize = kwargs.get('normalize', True)
+    marker = kwargs.get('marker')
+    normalization_dict = kwargs.get('normalization_dict')
+    if normalize:
+        if marker in normalization_dict.keys():
+            norm_factor = normalization_dict[marker]
+        else:
+            print("Norm_factor not found for marker {}".format(marker))
```
ngreenwald:
```suggestion
            print("Norm_factor not found for marker {}, calculating directly from the image. ".format(marker))
```

ngreenwald (changes requested):
Looks good, just a few small ones

cell_classification/application.py
JLrumberger: We don't want to normalize the binary mask in channel 2.

cell_classification/application_test.py
JLrumberger: Agree!

ngreenwald (approved):
Looks good!

PR #42, opened 2022-12-16
Added SimpleTFRecords class for handling MSK data

### LINKED ISSUE #41: Simple version of SegmentationTFRecords that expects image & label instead of conversion matrix
### Instructions

Implement a new class with all basic functionality to prepare .tfrecords datasets without assuming a `conversion_matrix` as input.

**Relevant background**

The MSK dataset comes without segmentation masks, thus we needed to segment the data with mesmer. To assign positive/negative classes for a marker to the cell segments, we use their cell table to calculate x: the percentage of positive cells for a given marker and then assign marker positive to the brightest x% of cells based on our segmentation. Thus this class should work without `conversion_matrix` and get's a cell_table that contains a column $marker+"_gt" where marker positivity is stored

**Design overview**

This is really tricky. We have two choices:
1. The `SegmentationTFRecords` class we got at the moment will inherit from a new class and implement some more functions to cope with conversion_matrix based activity map generation.
2. We go the other way round and build a child class that inherits from `SegmentationTFRecords` and just needs to implement `__init__` and `get_marker_activity` and a new class function called `check_additional_inputs`. All checks on input that assume a conversion_matrix will then be implemented in `SegmentationTFRecords.check_additional_inputs` while this function will be empty in the new child class.

The 1st option would be a bit cleaner but more code especially in the tests needs to be changed. The 2nd option only requires minimal code changes and would be considerably faster to implement. The general problem here is that in object oriented design you normally have the most abstract class as the root and everything that inherits from it will be more specialized and less abstract. `SegmentationTFRecords` would implement `check_additional_inputs` whereas it's empty in our new class, this means that the former is more specialized than the latter and thus `SegmentationTFRecords` should inherit from our new class. But since we will only use this code for the MSK dataset and don't run it very often, I'd still go with option 2. because it's faster to implement, even though it's not clean design from an OOD standpoint.

**Code mockup**

class SimpleTFRecords(SegmentationTFRecords):
    def __init__():
        # init everything except conversion_matrix

    def get_marker_activity(self, sample_name, marker):
        # get marker activity from activity_df[$marker+"_gt"]

   def check_additional_input():
       pass

class SegmentationTFRecords:

   def check_additional_input():
        # check conversion_matrix

**Required inputs**

All inputs from `SegmentationTFRecords` except `conversion_matrix_path` and in addition `gt_suffix` which specifies the suffix in column name $marker+gt_suffix used to load the GT

**Output files**

.tfrecord dataset

**Timeline**
- [x] A couple days
- [ ] A week

Estimated date when a fully implemented version will be ready for review: 12/16/2022

Estimated date when the finalized project will be merged in: 12/20/2022

JLrumberger:

This PR closes #41 by adding a subclass `SimpleTFRecords` that inherits from `SegmentationTFRecords`. Instead of requiring a `conversion_matrix`, it requires a cell_table that contains column $marker+`self.gt_suffix` that contains 0s/1s/2s labels for each cell segment. 

I subclasses `SegmentationTFRecords`, implemented new functions for `__init__`, `get_image` and `get_inst_binary_masks`. In `SegmentationTFRecords.load_and_check_input` everything that is specific to `conversion_matrix`-based GT-generation is moved to `check_additional_inputs`.

**Remaining issues**

None

### DIFF
--- a/cell_classification/prepare_data_script_MSK_colon.py
+++ b/cell_classification/prepare_data_script_MSK_colon.py
@@ -1,0 +1,37 @@
+import os
+from simple_data_prep import SimpleTFRecords
+os.environ["CUDA_VISIBLE_DEVICES"] = "-1"
+
+
+def naming_convention(fname):
+    return os.path.join(
+        "C:/Users/lorenz/Desktop/angelo_lab/data/MSKCC_colon/segmentation",
+        fname + "feature_0.ome.tif"
+    )
+
+
+data_prep = SimpleTFRecords(
+    data_dir=os.path.normpath(
+        "C:/Users/lorenz/Desktop/angelo_lab/data/MSKCC_colon/raw_structured"
+    ),
+    cell_table_path=os.path.normpath(
+        "C:/Users/lorenz/Desktop/angelo_lab/data/MSKCC_colon/cell_table.csv"
+    ),
+    imaging_platform="Vectra",
+    dataset="MSK_colon",
+    tile_size=[256, 256],
+    stride=[240, 240],
+    tf_record_path=os.path.normpath("C:/Users/lorenz/Desktop/angelo_lab/data/MSKCC_colon"),
+    normalization_quantile=0.999,
+    selected_markers=["CD3", "CD8", "Foxp3", "ICOS", "panCK+CK7+CAM5.2", "PD-L1"],
+    sample_key="fov",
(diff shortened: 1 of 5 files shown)


ngreenwald (approved):
Looks good, I agree that changing the minimal amount of code makes sense for a use-case that only matters for these two datasets

PR #50, opened 2023-02-25
Adding more control and train/val/test set composition

### LINKED ISSUE #49: Adding more control for validation and test dataset composition
### Instructions
Add a class method to `ModelBuilder` that filters sets of FOV names from the `tfrecord` into the respective test and validation datasets. 

**Relevant background**

Instead of taking the first x tiles as test and validation data, we want to get more control over the composition of the datasets by explicitly using lists of FOV names to construct them.

**Design overview**

Implement function `ModelBuilder.filter_fovs(self, dataset, fov_list, positive_list)`, that takes a dataset, a list of FOVs and a boolean indicating if the fov_list is a positive or negative list, meaning if the respective fovs should be filtered out of the dataset or filtered into a dataset that gets returned. The function returns one dataset that only contains samples whose FOVs are in `fov_list` if `positive_list=True` and one dataset that contains all samples except the ones whose FOVs are in `fov_list` if `positive_list=False`.

**Code mockup**

```
def filter_fovs(self, dataset, fov_list, positive_list, fov_key="fov"):
    dataset_tmp = copy(dataset)
    if positive_list:
        def predicate(fov):
            return fov in fov_list
    if not positive_list:
        def predicate(fov):
            return fov not in fov_list

    dataset_tmp = dataset_tmp.filter(lambda example: tf.py_function(predicate, [example[fov_key]], tf.bool))
    return dataset_tmp
```

**Required inputs**
.tfrecord dataset that holds `fov_key` as key and list `fov_list` that is stored as .json and will be loaded in the init.

**Output files**

Loaded tfrecord dataset that holds samples from the specified FOVs.

**Timeline**
- [x] A couple days
- [ ] A week

JLrumberger:

This PR closes #49 and adds function `ModelBuilder.fov_filter` to filter datasets based on lists that contain the fov names. The actual train/test/validation splits are saved individually for each dataset in e.g. `configs/tonic_split.json` and loaded in `ModelBuilder.prep_data`.

If we supply `params["data_splits"] = ["path_to_/dataset_split.json"]` to `ModelBuilder` or `PromixNaive` objects, calling class method `prep_data` loads the .json files into dicts with three keys each (`train`, `validation` and `test`). The associated values are lists of FOVs that should be used for train / validation / test datasets. To do this, class function `ModelBuilder.fov_filter(self, dataset, fov_list, fov_key)` is called to construct the train / validation and test datasets. `ModelBuilder.fov_filter` simply takes `fov_list`, constructs a predicate and maps it onto the full dataset to filter out all samples showing FOVs that are not in `fov_list`.

**Remaining issues**

None


cell_classification/model_builder.py
```diff
@@ -63,21 +63,42 @@ def prep_data(self):
...
+                )
+            ]
+            self.train_datasets = [
+                self.fov_filter(dataset, data_split["train"]) for dataset, data_split in zip(
+                    datasets, data_splits
+                )
+            ]
+        else:
+            self.validation_datasets = [
+                dataset.take(num_validation) for dataset, num_validation in zip(
```
ngreenwald, line 88: Where does the randomization happen? I see that the JSON files included properly randomized splits, so it's happening somewhere, but this just takes the first `x` entries, not a random `x`

ngreenwald (approved):
Looks great!

cell_classification/model_builder.py
JLrumberger: Yep, shuffling happens later. I did the train/val/test splits without shuffling, so that I have disjoint sets of FOVs in the dataset subsets. Shuffling the tiles before would mix FOVs in the subsets which I wanted to avoid.

ngreenwald: Right, but how are the FOVs themselves shuffled? i.e. not in alphabetical order?

JLrumberger: If you don't provide `data_splits`, FOVs are in alphabetical order and the train/val/test split via `dataset.take()` will put them into subsets like this `[a, b], [c, d], [e,f,g,h,...,z]`. If you provide `data_splits`, it is assumed that FOVs in `data_splits` are already shuffled. After datasets are split in one way or the other, the tiles in the train dataset are shuffled.

ngreenwald: Okay, so in order to generate the TONIC `data_splits`, which look like `[[TMA1, TMA9], [[TMA22, TMA3], [TMA18, TMA14]]`, you had to provide the datasplits? The existing logic, if supplied no data_splits, would have instead done `[[TMA1, TMA2], [TMA3, TMA4], [...TMA18, TMA19?]`

PR #59, opened 2023-04-20
Application notebooks

JLrumberger:

Added a notebook `predict.ipynb` in the style of the ark-analysis template notebooks that calculates a normalization dictionary, loads the classification model and does predictions on FOVs. The output predictions can be saved as images and plotted within the notebook and are written into a cell table that will be saved as `pred_cell_table.csv`.

In this PR, all notebook specific functions are still defined in the notebook. In future versions this will be put in a python file and imported in the notebook. 

**Remaining issues**

To import functions from the cell_classification folder, we need to add "../src" to path.


ngreenwald (approved):
Looks good. One you've sorted out the bug in the cluster assignment labels, I think it'll be worth going back and cleaning this up. It should also be more clear what the workflow is at that point. 

Then we can start having more people try it out.

PR #60, opened 2023-04-21
Added seeds to tests and changed conftest to work on my machine

JLrumberger:
Hi @srivarra,
I got the tests to run on my machine and here's a PR that has the necessary changes. I'd be happy to get some feedback from you.

Purpose of this PR is to make the tests run on my local device , without installing the codebase as a package. Installing the module is not desired, because it makes debugging more complicated. If installed, Python would import modules from the site-packages instead of the actuall code base, which prevents the breakpoints in other parts of the library from stopping code execution while debugging. To make it work without installing, I needed to set PYTHONPATH variables for the `tests` and the `src` folders. Also I got some errors, because the code used all available GPUs, thus I hide them for the tests.

I added `CUDA_VISIBLE_DEVICES="-1"` and sys paths in `conftest.py` so that all tests run on my local device without installing cell_classification as a library. I also added a few `np.random.seed(42)` in one of the test scripts to make tests more reproducible.

**Remaining issues**

I am not sure if changing `conftest.py` is the right way to get the tests to run on my machine. I could also add a .env file, but then I need to see how to add this .env to the workspace settings in my IDE. I went for changing `conftest.py` because it was the easiest way. Is there a best practice here?


JLrumberger:
Talked to @srivarra and tried to get around adding the path variables in conftest.py by doing `pip install -e .[test]` which installs the module in editable mode. This threw path errors when running `pytest -c tests`, so I reverted back to adding the paths in conftest. Since it doesn't break anything in the CI and let's the test run on my machine, I think we can merge it in.

tests/segmentation_data_prep_test.py
```diff
@@ -64,6 +64,7 @@ def prep_object_and_inputs(
 
 
 def prepare_conversion_matrix():
```
srivarra, line 66: A bit nitpicky, but I'd suggest passing in the `rng` fixture which lives in `conftest.py` and using the [Generator API](https://numpy.org/doc/stable/reference/random/generator.html) to these tests / test utility functions.

For example,

```python
def prepare_conversion_matrix(rng: np.random.Generator):
	rng.integers(low=0, high = 3, size=(6,4)).clip(0,1)
	...
```

This generates integers from the same seed, so it'll be reproducible on different systems (even though it only runs on linux for now).

srivarra (changes requested):
Looks good, it's kinda odd that it wasn't working without adding those path variables, but it's good that it's working. Just a _small_ biased suggestion.

tests/segmentation_data_prep_test.py
JLrumberger: That's a cool idea, but I think it doesn't work for this particular function since it does not have `test` prepended in the name, which is required for pytest to supply rng as an argument. I tried decorating the function with `@pytest.mark.usefixtures("rng")`, but that also didn't work. So I'd need to change the name of the function to `test_prepare_conversion_matrix` which also isn't a good solution, since the function is a helper function and does not contain tests. Also I reference the function ~10 times in the codebase and would need to change those references too.
Do you have an idea on how to make use of the pytest fixtures in functions that do not contain "test" in its name?

srivarra: Oh yeah, that's a good observation. I don't believe there is a way to pass fixtures to functions which are not tests nor other fixtures. Don't worry about it then!

srivarra (approved):

PR #65, opened 2023-08-15
Prediction notebook

### LINKED ISSUE #64: Clean up application notebooks
**Is your feature request related to a problem? Please describe.**
The current application notebooks have a bunch of functions defined within them, rather than imported. In addition, there are some manual steps (defining marker names) that should probably be handled automatically. There's also stuff about the google drive integration, example datasets, etc that was copied from `ark` that we should remove or update to be correct. 

All of this cleanup will be good to do before we start having people try out the pipeline so we can figure out what's working and what isn't.

JLrumberger:

This PR closes #64 and adds the class `Nimbus` which extends the class `deepcell.applications.Application` and gives an interface for all things inference. The class first checks its inputs, loads the model, then prepares the normalization dictionary and does predictions. In addition, the class `NimbusViewer` implements a viewer widget, than can be used to scroll through FOVs and show input channels and output predictions side by side for visual inspection of the results. Both classes are used in the `predict.ipynb` to first predict model confidence scores and then inspect them.

1. I added several functions to `inference.py` that implement a functional interface for inference. The following functions are there
- `prepare_normalization_dict`: iterates through FoVs and calculates the normalization quantile. Has the option to run parallel via joblib.
- `prepare_input_data`: turns the segmentation map to a binary representation with eroded edges
- `segment_mean`: takes instance segmentation map and another image and returns mean per cell expression profiles.
- `test_time_aug`: Does inference over a batch of flipped and 90°-rotated images and returns the average over these transformations. This is known to improve results slightly.
- `predict_fovs`: iterates through fovs and predicts
 
2. I added the class `Nimbus` that extends `deepcell.applications.Application` and gives a class-based interface to the above mentioned inference functions.
- `check_input`: Checks the input for correctness
-  `initialize_model`: Initializes ML model and loads the model checkpoint 
- `prepare_normalization_dict`: does what it says :)
- `predict_fovs`: iteratives through fovs, predicts and stores them

3. `NimbusViewer` is a class that implements an ipywidget to view multiplexed channels and predicted model confidence scores side-by-side and allows the user to choose which FOV and channels to look at.

**Remaining issues**
`NimbusViewer` does not have tests yet. I'll look into the Pixie repo to see how widgets can be tested.


templates/predict.ipynb
```diff
@@ -21,34 +21,27 @@
    ],
```
ngreenwald: <p>Here we should spell out again what the default assumption is: that all segmentation TIFs are in a single directory with a specified suffix. Then if people want to change things with segmentation_naming_convention, we can tell them to copy it in and modify. But it'll make the notebook cleaner if we default to not having that in there. </p><p> </p><p>Also, I think <code>ark</code> got rid of feature_0 and uses a more intuitive name now, you can double check.  </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/cell_classification/pull/65/discussion/#comment-1298904947'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='5'/>

ngreenwald: <p>How long does this take for a single 2048x2048 FOV?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/cell_classification/pull/65/discussion/#comment-1298904948'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

ngreenwald: <p>How long does this take to make predictions for a single 2048x2048x1 image? 2048x2048x15?</p><p> </p><p>What about on a CPU workstation? i.e. should we tell people they definitely need a GPU?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/cell_classification/pull/65/discussion/#comment-1298904950'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='11'/>

ngreenwald: <p>What do these plots end up looking like?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/angelolab/cell_classification/pull/65/discussion/#comment-1298904952'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='RESOLVED' data-cellIndex='13'/>

ngreenwald (changes requested):
Notebook is looking quite nice! Just a couple suggestions for making it simpler, and questions about performance

templates/predict.ipynb
JLrumberger: Plots look like this:
NimbusViewerWidget

JLrumberger: RTX 3090 24GB on a workstation: 54 seconds for 1 FoV (2048x2048x35), so ~1.5sec for 1 image with test time augmentations (TTA)
RTX3050 4GB on a notebook: 150seconds for one FoV or ~4.3sec per image with TTA
i7 11800H 8-core CPU on notebook: 44 minutes for one FoV or 75sec per image with TTA. Without TTA, 3.3 sec per image 115 sec per FoV
I'll add text, saying that CPU inference is okay without TTA, but GPU inference is always preferred.

JLrumberger: On my notebook i7 11800H 8-core and SSD (3.5G/sec read): with multiprocessing=True 0.25 sec/FOV, multiprocessing=False 3.9sec/FOV. So this is blazing fast for machines with SSD, but probably really slow for old machines with HDD (e.g. 80-160mb/sec read).

ngreenwald: <p>Okay got it. I think setting this default closer to 20 or 40 FOVs would be better. I guess you could also look and see what the sampling distribution looks like. Your call</p>

ngreenwald: <p>TTA makes it 25x slower? 3.3 sec -&gt; 75 seconds?</p><p> </p><p>Yeah I agree, and it's great that it'll work decently on CPU</p>

ngreenwald: <p>Looks great!</p>

JLrumberger: That's a bit extreme, the difference should be 6-fold, because TTA averages over 6 different views of the data. I guess the 25-fold increase is due to CPU throttle, my laptop got pretty hot from running these experiments.

JLrumberger: Good call, I'll look into it.

JLrumberger:
I'll add a few more tests for the viewer widget tomorrow and then will merge this PR before the workshop. Is that okay @ngreenwald ?

PR #73, opened 2023-10-03
Vanilla unet with valid/same/mirror padding

### LINKED ISSUE #68: Unexplainable "tiling shadows" in prediction masks
Dear Lorenz, der devteam,

First of all thanks a lot for this cool and exciting new tool! Looking forward to reading the preprint soon.
I tried to run Nimbus on some of my melanoma samples and ran into some peculiarities. To me it seems that while the predictions for positivity are spot-on for some markers, I see weird square-shaped shadows on some other channels with the dimensions of 1/3rd the fov-size (see image, these FOVs are 800x800µm). Further, for some channels (e.g. CD86 in this example), it appears to me that antibody-mediated background signal is overemphazised a bit when comparing the input image (with "true" positive cells in the fov centre) and the corresponding prediction. Do you have ideas of where the issue may lie here? Would be happy to get your insights on this. I ran nimbus on the most recent version and with default args for all functions except input_shape for Nimbus(), which i put to [512,512] as I am running the package on my laptop.

Thanks a lot for your help!

Best wishes,
Sven

nimbus

JLrumberger:

This PR closes #68 by using a vanilla unet and translation equivariant tile & stitch inference with valid convolutions.

I adapted the code from https://github.com/jakeret/unet and added batchnorm layers and mirror-padding.

**Remaining issues**
None

### DIFF
--- a/src/cell_classification/model_builder.py
+++ b/src/cell_classification/model_builder.py
@@ -2,7 +2,7 @@ import argparse
 import json
 import os
 from time import time
-
+from cell_classification.unet import build_model
 import h5py
 import numpy as np
 import pandas as pd
@@ -179,6 +179,12 @@ class ModelBuilder:
                         1, (3, 3), input_shape=self.params["input_shape"], padding="same",
                         name="semantic_head", activation="sigmoid", data_format="channels_last",
                     )]
+            )
+        elif self.params["backbone"] == "VanillaUNet":
+            self.model = build_model(
+                nx=self.params["input_shape"][0], ny=self.params["input_shape"][1],
+                channels=self.params["input_shape"][2], num_classes=self.params["classes"],
+                data_format="channels_last", padding="REFLECT"
             )
         else:
             self.model = PanopticNet(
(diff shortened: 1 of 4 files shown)


ngreenwald (approved):
So what was it about the previous backbone that was causing these issues?

JLrumberger:
My guess is that padding in the convolutions caused these issues. When you use padding with convolutions, they are not shift-equivariant anymore, meaning that they can learn to incorporate absolute position of features into their predictions, thereby producing these tiling shadows mentioned in #68. The previous backbone was a ResNet50, when you turn off padding in the conv-layers the intermediate filter maps get smaller by 2 pixels in X and Y direction after each filter operation, thus the output will become really small (assuming a constant input image size).
My plan is to use a regular U-Net (with less layers in the encoder than the other backbone), train it with reflective-padding. For inference, we'll use valid (i.e. no) padding, then it'll be provably translation equivariant and tiling shadows will be gone.

An alternative would be to optimize for equivariance by predicting two representations of the input data, one with shifted input and one without shifted input and enforce pixel-wise consistency after applying these transformations on the output. The latter approach doesn't have guarantees, while for the former method the equivalence of doing tile & stitch inference and predicting the whole image at once has been proven.

PR #80, opened 2024-02-01
Wandb 

JLrumberger:

In this PR we add a residual unet with mirror padding as a new backbone architecture and wandb logging of validation metrics to the training pipeline.

I replaced the tensorboard logging functionality in ModelBuilder and PromixNaive with wandb logging and changed the vanilla unet architecture to use resnet style skip connections in the conv blocks. In addition I threw out a couple of tests from the deepcell PanopticNet, because they threw assertions without me changing any code or dependencies and I couldn't resolve them within a couple of hours.

**Remaining issues**

None


ngreenwald (approved):
Looks good! This is the one that had best performance in the benchmarking?

JLrumberger:
> Looks good! This is the one that had best performance in the benchmarking?

Yep

